PR #2, opened 2025-10-25
docs: Phase 4 Progress Update - Token Refresh, CORS Fixes, and Upload UI

leixiaoyu:
## Summary

This PR updates project documentation to reflect Phase 4 progress (now at 75% completion) and documents recent feature implementations including automatic token refresh, API Gateway CORS fixes, and file upload UI components.

### Documentation Updates
- **README.md**: Updated overall progress to 25%, Phase 4 status to 75%
- **PROGRESS.md**: Added comprehensive Phase 4 achievements section with detailed metrics

#### 1. API Gateway CORS Enhancements
- Added Gateway Responses for error codes (401, 403, 400, 5XX)
- CORS headers now present on all error responses
- Fixed authentication error CORS blocking
- Production deployment verified

#### 2. Automatic Token Refresh
- Response interceptor with 401 error handling
- Request queuing during token refresh
- Prevents multiple concurrent refresh calls
- Retry mechanism with `_retry` flag
- Fallback to logout if refresh fails
- Comprehensive test coverage (8 test scenarios)

#### 3. User Interface Enhancements
- App bar with logout button
- User email display in header
- Navigation after logout
- Material-UI styled components
- 21 comprehensive UI tests

#### 4. File Upload UI Components
- FileUploadForm component with drag-and-drop
- NewTranslationPage with upload integration
- Progress tracking and upload status display
- File validation (size, type, format)
- Upload service abstraction layer
- Comprehensive test coverage

### Test Coverage
- **New Tests**: 29+ tests added in Phase 4
- **Total Tests**: 290+ tests passing (252+ frontend + 38 infrastructure)
- **Coverage**: Maintained >90% overall test coverage
- **Test Files Added**:
  - `api.refresh.test.ts` (8 test scenarios for token refresh)
  - `NewTranslationPage.test.tsx` (21 test scenarios for UI)
  - `FileUploadForm.test.tsx` (upload component tests)
  - `uploadService.test.ts` (service layer tests)

### Files Changed (12 files, 2,596 insertions)
- `README.md` - Updated status and progress
- `PROGRESS.md` - Added Phase 4 section
- `backend/infrastructure/lib/lfmt-infrastructure-stack.ts` - Gateway Responses
- `frontend/src/utils/api.ts` - Token refresh interceptor
- `frontend/src/pages/NewTranslationPage.tsx` - New page with upload UI
- `frontend/src/components/Translation/FileUploadForm.tsx` - Upload component
[...]

PR #126, opened 2026-04-14
fix: Backend/infrastructure — issues #17, #20, #16, #44, #40

leixiaoyu:
## Summary
Backend and infrastructure improvements.

### Fixed
- **#17** — API key secret: Created translationApiKeySecret in CDK stack, grants read access to translation Lambda
- **#20** — File validation constants: Centralized in shared-types/src/validation.ts, single source of truth
- **#16** — S3 lifecycle contradiction: Removed misleading comment, clarified STANDARD→IA→GLACIER transition path
- **#44** — MAX_CONCURRENCY: Already uses CDK context (this.node.tryGetContext), supports env-specific tuning
- **#40** — RateLimitError: Custom exception in rateLimiting.ts, used by DistributedRateLimiter.acquire()

### Files Changed
- backend/infrastructure/lib/lfmt-infrastructure-stack.ts (secrets, lifecycle)
- shared-types/src/validation.ts (consolidated constants)
- backend/functions/translateChunk.ts (RateLimitError handling)

### Testing
- Backend tests: passed

## Related PR absorbed

This PR now absorbs the 2 test-mock fix commits from **PR #129** (sibling branch `fix/pr126-distributed-rate-limiter-test`):
- `641aec6` fix: use callsFake instead of resolvesOnce for DynamoDB mocks (ported from `f94d458`)
- `4a87fac` fix: re-throw RateLimitError instead of catching as DynamoDB error (ported from `85d7cdc`)

Additionally, one more test-mock fix was needed for the same class of issue:
- `0264a96` fix(test): update translateChunk rate-limit mock to throw RateLimitError

With those integrated, all backend unit tests pass (407/407). Once this PR merges, PR #129 should be closed as redundant.

PR #167, opened 2026-04-30
fix(translateChunk): remove rate-limiter handler param that caused #150 TypeError

### LINKED ISSUE #150: [BUG-CRITICAL][P0] lfmt-translate-chunk-LfmtPocDev Lambda is stale — TypeError: i.acquire is not a function on every invocation since 2026-03-19

leixiaoyu:
## Summary

Closes #150

- **Bug**: Every invocation of `lfmt-translate-chunk-LfmtPocDev` threw `TypeError: i.acquire is not a function` (bundle `index.js:3358:11660`), failing 100% of translations since 2026-03-19.
- **Root cause**: `handler(event, _rateLimiter?)` — AWS Lambda always calls handlers as `handler(event, context)`, so the Lambda context object landed in `_rateLimiter`. Because context is truthy, the `if (_rateLimiter) { rateLimiter = _rateLimiter; }` branch always fired, assigning the context as the rate limiter. `rateLimiter.acquire()` then crashed on every call.
- **Why PR #126 missed it**: PR #126 migrated the `acquire()` return-value API to a throw-based API and updated the test mock to pass via the second handler argument — but never questioned the signature itself. Tests passed because they pass a real mock object; production always received the Lambda context instead.

## The hunt

1. Downloaded the deployed bundle (`CodeSha256: O6cbyI4s68MckTGVwskPp3fhAqWFmPQ6g8u4I9rRNFc=`, `LastModified: 2026-04-27T23:26:14`) and extracted 200 chars around `index.js:3358:11660`:
   ```
   e?i=e:(ac||(ac=new Kl({...})),i=ac) ... await i.acquire(c,"tpm")
   ```
   `e` = second handler argument = Lambda context. `i = e` = context. `i.acquire` = undefined. Case closed.
2. Confirmed the source has exactly one `.acquire(` call (`translateChunk.ts:154`) — no missed call sites from PR #126. The bug is the handler signature, not a missed migration.

## Fix

- Removed `_rateLimiter` from the handler signature (`handler.length` is now 1).
- Added `setRateLimiterForTesting(rateLimiter)` export alongside `resetClients()` as the sole DI path for unit tests.
- Updated the existing rate-limit mock injection test to use `setRateLimiterForTesting()` instead of `handler(event, mockRateLimiter)`.

## Regression tests added

Two new tests in `REGRESSION #150: handler signature must not accept a rate-limiter parameter`:
- `"handler must accept exactly one parameter so Lambda context is never mistaken for a rate limiter"` — asserts `handler.length === 1`.
- `"setRateLimiterForTesting() must be the only DI path and must be cleared by resetClients()"` — injects a mock limiter, verifies `acquire()` is called and rate-limit error propagates, then confirms teardown clears the injection.

All 31 `translateChunk` tests pass; 0 ESLint errors; Prettier clean.

## Verification plan

After merge → next push redeploys Lambda → manually start a translation against dev (`https://d39xcun7144jgl.cloudfront.net`) → verify `translationStatus` advances to `COMPLETED` (not stuck on `i.acquire` TypeError in CloudWatch, and not silently flipped to `TRANSLATION_FAILED`).

### DIFF
--- a/backend/functions/translation/__tests__/translateChunk.test.ts
+++ b/backend/functions/translation/__tests__/translateChunk.test.ts
@@ -12,7 +12,12 @@ import { DynamoDBClient, GetItemCommand, UpdateItemCommand } from '@aws-sdk/clie
 import { S3Client, GetObjectCommand, PutObjectCommand } from '@aws-sdk/client-s3';
 import { SecretsManagerClient, GetSecretValueCommand } from '@aws-sdk/client-secrets-manager';
 import { Readable } from 'stream';
-import { handler, TranslateChunkEvent, resetClients } from '../translateChunk';
+import {
+  handler,
+  TranslateChunkEvent,
+  resetClients,
+  setRateLimiterForTesting,
+} from '../translateChunk';
 import { sdkStreamMixin } from '@smithy/util-stream';
 import { RateLimitError, RateLimitType } from '../../shared/types/rateLimiting';
 
@@ -968,6 +973,12 @@ describe('translateChunk Lambda', () => {
         ),
       } as any;
 
+      // Inject via setRateLimiterForTesting() — NOT as a handler parameter.
+      // Passing a rate limiter as handler(event, limiter) is what caused issue #150:
+      // Lambda passes context as the second argument, which was mistakenly used as
+      // the limiter, leading to "i.acquire is not a function" on every invocation.
+      setRateLimiterForTesting(mockRateLimiter);
+
       const event: TranslateChunkEvent = {
         jobId: 'job-123',
         userId: 'user-123',
@@ -975,8 +986,7 @@ describe('translateChunk Lambda', () => {
         targetLanguage: 'es',
       };
 
-      // Pass mock rate limiter to handler
-      const result = await handler(event, mockRateLimiter);
+      const result = await handler(event);
 
       expect(result.success).toBe(false);
       expect(result.error).toContain('Rate limit exceeded');
@@ -987,6 +997,70 @@ describe('translateChunk Lambda', () => {
 
       // Verify rate limiter was called
       expect(mockRateLimiter.acquire).toHaveBeenCalled();
+    });
+  });
+
+  describe('REGRESSION #150: handler signature must not accept a rate-limiter parameter', () => {
+    // Root cause of issue #150: the handler previously had a second optional
+    // parameter `_rateLimiter?: DistributedRateLimiter`.  AWS Lambda always
+    // calls handlers as handler(event, context), so the Lambda context object
+    // was silently assigned to `rateLimiter`.  Because context is truthy the
+    // DI branch was taken and every invocation threw
+    // "TypeError: i.acquire is not a function" (bundle line 3358 col 11660).
+
+    it('handler must accept exactly one parameter so Lambda context is never mistaken for a rate limiter', () => {
+      // handler.length is the number of *declared* parameters (not counting
+      // rest params or parameters with defaults).
+      expect(handler.length).toBe(1);
+    });
+
+    it('setRateLimiterForTesting() must be the only DI path and must be cleared by resetClients()', async () => {
+      const mockRateLimiter = {
+        acquire: jest.fn().mockRejectedValue(
+          new RateLimitError({
+            tokensNeeded: 500,
+            tokensAvailable: 0,
+            retryAfterMs: 1000,
+            limitType: RateLimitType.TPM,
+          })
+        ),
+      } as any;
+
+      dynamoMock.on(GetItemCommand).resolves({
+        Item: createMockJob({
+          jobId: 'job-reg150',
+          userId: 'user-123',
+          status: 'CHUNKED',
+          totalChunks: 1,
+        }),
+      } as any);
+
+      s3Mock.on(GetObjectCommand).resolves({
+        Body: createMockStream(
+          JSON.stringify({ primaryContent: 'Regression test', chunkId: 'chunk-0' })
+        ),
+      } as any);
+
+      // Inject the mock limiter via the dedicated setter
+      setRateLimiterForTesting(mockRateLimiter);
+
+      const event: TranslateChunkEvent = {
+        jobId: 'job-reg150',
+        userId: 'user-123',
+        chunkIndex: 0,
+        targetLanguage: 'es',
+      };
+
+      const result = await handler(event);
+
+      // Mock was used — rate-limit error propagated correctly
+      expect(mockRateLimiter.acquire).toHaveBeenCalled();
+      expect(result.success).toBe(false);
+      expect(result.retryable).toBe(true);
+
+      // After resetClients(), the injected mock must be cleared
+      resetClients();
+      expect(mockRateLimiter.acquire).toHaveBeenCalledTimes(1); // no additional calls after reset
     });
   });
 
--- a/backend/functions/translation/__tests__/translateChunk.test.ts
+++ b/backend/functions/translation/__tests__/translateChunk.test.ts
@@ -1058,9 +1058,165 @@ describe('translateChunk Lambda', () => {
       expect(result.success).toBe(false);
       expect(result.retryable).toBe(true);
 
-      // After resetClients(), the injected mock must be cleared
+      // Snapshot the call count *before* the reset so we can later prove
+      // that a fresh handler invocation did NOT use the injected mock.
+      const callCountBeforeReset = mockRateLimiter.acquire.mock.calls.length;
+
+      // After resetClients(), the injected mock must be cleared.  The
+      // previous version of this test asserted toHaveBeenCalledTimes(1)
+      // immediately after resetClients(), but the handler was never
+      // re-invoked, so the assertion was tautologically true regardless
+      // of whether the injection had actually been cleared.  We now
+      // re-invoke the handler post-reset and assert the mock's call
+      // count did NOT increase — proving the singleton was actually
+      // re-initialized fresh and the test injection is gone.
       resetClients();
-      expect(mockRateLimiter.acquire).toHaveBeenCalledTimes(1); // no additional calls after reset
+
+      // Re-stub DynamoDB/S3 because resetClients() does not touch the
+      // SDK mocks and we want the second invocation to reach the rate
+      // limiter (rather than failing earlier on missing job data).
+      dynamoMock.on(GetItemCommand).resolves({
+        Item: createMockJob({
+          jobId: 'job-reg150',
+          userId: 'user-123',
+          status: 'CHUNKED',
+          totalChunks: 1,
+        }),
+      } as any);
+      s3Mock.on(GetObjectCommand).resolves({
+        Body: createMockStream(
+          JSON.stringify({ primaryContent: 'Regression test', chunkId: 'chunk-0' })
+        ),
+      } as any);
+
+      // Second invocation post-reset.  This will use the real, lazy-
+      // initialized DistributedRateLimiter (talking to the mocked
+      // DynamoDB).  We don't care whether this call succeeds or fails;
+      // we only care that (a) it does not throw "acquire is not a
+      // function" (which would mean the mock was never cleared), and
+      // (b) the injected mock's call count is unchanged.
+      let secondInvocationError: unknown = undefined;
+      try {
+        await handler(event);
+      } catch (err) {
+        secondInvocationError = err;
+      }
+
+      // (a) If anything escaped the handler, it must NOT be the
+      //     original #150 TypeError shape.
+      if (secondInvocationError instanceof TypeError) {
+        expect(secondInvocationError.message).not.toMatch(/acquire is not a function/);
+      }
+
+      // (b) The injected mock must not have been touched by the second
+      //     invocation — proving resetClients() actually cleared the
+      //     test injection rather than leaking it across invocations.
+      expect(mockRateLimiter.acquire.mock.calls.length).toBe(callCountBeforeReset);
+    });
+
+    it('handler invoked with a Lambda context object must not treat it as a rate limiter', async () => {
+      // INTEGRATION-STYLE GUARD against the original #150 failure mode.
+      //
+      // The sibling test asserts handler.length === 1, which is a useful
+      // structural invariant — but it is NOT sufficient on its own.  A
+      // future contributor who re-introduces a default-value second
+      // parameter, e.g.
+      //
+      //     export async function handler(
+      //       event: TranslateChunkEvent,
+      //       _ctx: DistributedRateLimiter | undefined = undefined,
+      //     ) { ... }
+      //
+      // would still satisfy handler.length === 1, because parameters
+      // with default values do not count toward Function.prototype.length.
+      // The structural test would silently pass while the original bug
+      // returns: AWS Lambda would pass `context` as the second argument,
+      // _ctx would be truthy, and any subsequent `_ctx.acquire(...)` call
+      // would throw "TypeError: i.acquire is not a function" — exactly
+      // the production failure that PR #167 fixes.
+      //
+      // This test directly invokes the handler the way AWS Lambda does
+      // (handler(event, context)) using a Lambda-context-shaped object
+      // that deliberately has NO `.acquire` method.  If any future
+      // regression causes the handler to treat that context as the rate
+      // limiter, the next acquire() call will throw the canonical
+      // "acquire is not a function" TypeError and this test will fail.
+
+      // Lambda-context-shaped mock.  Critically: NO `acquire` method.
+      const mockLambdaContext = {
+        awsRequestId: 'test-req-id',
+        functionName: 'lfmt-translate-chunk-LfmtPocDev',
+        functionVersion: '$LATEST',
+        invokedFunctionArn:
+          'arn:aws:lambda:us-east-1:000000000000:function:lfmt-translate-chunk-LfmtPocDev',
+        memoryLimitInMB: '512',
+        getRemainingTimeInMillis: () => 30000,
+        callbackWaitsForEmptyEventLoop: true,
+        logGroupName: '/aws/lambda/lfmt-translate-chunk-LfmtPocDev',
+        logStreamName: '2026/04/28/[$LATEST]abcdef0123456789',
+      };
+
+      // Wire up minimal DynamoDB / S3 mocks so the handler reaches the
+      // rate-limit acquire() call (which is where the original bug
+      // manifested).  We do NOT inject a test rate limiter — the goal
+      // is to verify that the handler ignores the second argument
+      // entirely and uses its singleton (real or test-overridden) path.
+      dynamoMock.on(GetItemCommand).resolves({
+        Item: createMockJob({
+          jobId: 'job-reg150-ctx',
+          userId: 'user-123',
+          status: 'CHUNKED',
+          totalChunks: 1,
+        }),
+      } as any);
+      s3Mock.on(GetObjectCommand).resolves({
+        Body: createMockStream(
+          JSON.stringify({ primaryContent: 'Lambda context regression test', chunkId: 'chunk-0' })
+        ),
+      } as any);
+      s3Mock.on(PutObjectCommand).resolves({} as any);
+      dynamoMock.on(UpdateItemCommand).resolves({} as any);
+
+      // Inject a mock limiter so the test does not depend on the real
+      // distributed rate limiter's DynamoDB behavior.  If the handler
+      // were to (incorrectly) treat the Lambda context as the limiter,
+      // it would shadow this injection and the next acquire() call
+      // would throw "acquire is not a function" on the context object.
+      const mockRateLimiter = {
+        acquire: jest.fn().mockResolvedValue({ allowed: true, tokensRemaining: 100 }),
+      } as any;
+      setRateLimiterForTesting(mockRateLimiter);
+
+      const event: TranslateChunkEvent = {
+        jobId: 'job-reg150-ctx',
+        userId: 'user-123',
+        chunkIndex: 0,
+        targetLanguage: 'es',
+      };
+
+      // Invoke the handler the way AWS Lambda does: handler(event, context).
+      // The handler must accept this without ever calling context.acquire().
+      let invocationError: unknown = undefined;
+      try {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        await (handler as any)(event, mockLambdaContext);
+      } catch (err) {
+        invocationError = err;
+      }
+
+      // Primary assertion: whatever happened, it MUST NOT be the
+      // original #150 failure shape.  This is the canonical regression
+      // signature that PR #167 was created to fix.
+      if (invocationError instanceof TypeError) {
+        expect(invocationError.message).not.toMatch(/acquire is not a function/);
+      }
+
+      // Secondary assertion: the injected mock limiter (the real DI
+      // path) was used — proving the handler did NOT mistake the
+      // Lambda context for a rate limiter.  If the bug regressed, the
+      // context object would shadow the injection and this would never
+      // be called (the handler would TypeError first).
+      expect(mockRateLimiter.acquire).toHaveBeenCalled();
     });
   });
 
--- a/backend/functions/translation/translateChunk.ts
+++ b/backend/functions/translation/translateChunk.ts
@@ -38,12 +38,26 @@ const RATE_LIMIT_BUCKETS_TABLE = getRequiredEnv('RATE_LIMIT_BUCKETS_TABLE');
 let geminiClient: GeminiClient | null = null;
 let distributedRateLimiter: DistributedRateLimiter | null = null;
 
+// Test-only override: set via setRateLimiterForTesting(); cleared by resetClients()
+let _testRateLimiter: DistributedRateLimiter | null = null;
+
 /**
  * Reset singleton instances (for testing)
  */
 export function resetClients(): void {
   geminiClient = null;
   distributedRateLimiter = null;
+  _testRateLimiter = null;
+}
+
+/**
+ * Override the rate limiter singleton for unit testing.
+ * Must NOT be called from production code — the handler no longer accepts
+ * a rateLimiter parameter because Lambda passes `context` as the second
+ * argument, which would previously be treated as the limiter.
+ */
+export function setRateLimiterForTesting(rateLimiter: DistributedRateLimiter): void {
+  _testRateLimiter = rateLimiter;
 }
 
 /**
@@ -74,14 +88,18 @@ export interface TranslateChunkResponse {
 }
 
 /**
- * Lambda handler for translating a single chunk
+ * Lambda handler for translating a single chunk.
+ *
+ * IMPORTANT: This handler intentionally accepts only one parameter (event).
+ * AWS Lambda invokes handlers as handler(event, context) — if a second
+ * parameter were present, the Lambda context object would be received there
+ * and mistakenly used as the rate limiter, causing "i.acquire is not a
+ * function" (issue #150). Use setRateLimiterForTesting() to inject a mock
+ * limiter in unit tests instead.
+ *
  * @param event - Translation event
- * @param _rateLimiter - Optional rate limiter for testing (uses singleton if not provided)
  */
-export const handler = async (
-  event: TranslateChunkEvent,
-  _rateLimiter?: DistributedRateLimiter
-): Promise<TranslateChunkResponse> => {
+export const handler = async (event: TranslateChunkEvent): Promise<TranslateChunkResponse> => {
   const startTime = Date.now();
 
   logger.info('Starting chunk translation', {
@@ -101,10 +119,10 @@ export const handler = async (
       await geminiClient.initialize();
     }
 
-    // Use provided rate limiter or initialize singleton
+    // Use test-injected rate limiter (unit tests only) or initialize singleton
     let rateLimiter: DistributedRateLimiter;
-    if (_rateLimiter) {
-      rateLimiter = _rateLimiter;
+    if (_testRateLimiter) {
+      rateLimiter = _testRateLimiter;
     } else {
       if (!distributedRateLimiter) {
         distributedRateLimiter = new DistributedRateLimiter({


leixiaoyu:

Scope: PR #167 fix for translateChunk handler signature bug. Reviewed `backend/functions/translation/translateChunk.ts`, the test file, and grepped all handlers under `backend/functions/`.

**Bottom line: no critical or high-severity findings. The fix is sound and reduces risk relative to the prior signature. Two defense-in-depth suggestions below.**

### Critical
None.

### Recommendations
None of the findings rise to "must-fix before merge." Items below are hardening suggestions only.

### Suggestions

1. **Add a runtime guard to `setRateLimiterForTesting()` (defense in depth).**
   The new export is part of the module's public API with only a JSDoc warning. A future caller (e.g., a new Lambda that mistakenly imports the wrong symbol) could disable distributed rate limiting — which would let translations bypass the 5 RPM / 250K TPM / 25 RPD Gemini quotas, potentially causing API-key suspension or cost overruns. Suggested guard inside the setter:
   ```ts
   if (process.env.NODE_ENV === 'production' || process.env.AWS_EXECUTION_ENV) {
     throw new Error('setRateLimiterForTesting() must not be called in Lambda runtime');
   }
   ```
   `AWS_EXECUTION_ENV` is auto-set by Lambda (e.g. `AWS_Lambda_nodejs18.x`) and is the most reliable signal. This is belt-and-suspenders — no current caller exists, but the guard makes the misuse fail loud rather than silent.

2. **Optional: ESLint `no-restricted-imports` rule.** Add a rule that flags any import of `setRateLimiterForTesting` from outside `__tests__/` directories. Cheaper than the runtime guard and catches the mistake at build time.

### Positive findings

- **No other handlers share the anti-pattern.** Grepped every `export const handler` under `backend/functions/` (auth/*, jobs/*, chunking/*, translation/*). All other handlers correctly declare a single `event` parameter. The bug was isolated to `translateChunk.ts` — no undeployed copies of the same defect.
- **Module-level state is safely contained.** `_testRateLimiter` defaults to `null`, is only set by the new exported setter (which production code does not call), and is cleared by `resetClients()`. The test file calls `resetClients()` in `beforeEach` (line 89), so even a mid-test crash cannot leak injected state into the next test. Lambda warm containers are unaffected — production code never imports the test file.
- **No information leakage from the original bug.** The handler's catch block returns only `error.message` to the client (`translateChunk.ts:271-278`) and logs `error.stack` to CloudWatch server-side. The original `TypeError: i.acquire is not a function` exposed no Lambda context fields (`invokedFunctionArn`, `awsRequestId`, etc.) to API consumers. The bug was a 100% DoS for translations but had no confidentiality impact.
- **No new auth/authz, input-validation, or user-data surfaces.** This is pure runtime plumbing. `validateEvent()` and `userId` ownership checks (in `loadJob`) are unchanged.
- **Regression tests are well-targeted.** `handler.length === 1` is a structural invariant that will catch any future re-introduction of the second parameter — a strong, low-maintenance assertion.

### Severity summary

| Finding | Severity | OWASP |
|---|---|---|
| Missing runtime guard on test-only setter | Low (defense in depth) | A05:2021 Security Misconfiguration |
| Missing lint rule for test-only import | Informational | A05:2021 |

**Recommendation: approve. The two suggestions can be addressed in a follow-up PR if desired; they are not blocking.**

leixiaoyu:

### No Critical Issues

The fix is correct and the dangerous handler signature is gone. No blocking concerns — everything below is quality craft.

---

### Recommendations

**1. `_testRateLimiter` is never cleared between tests in the existing rate-limiter test block**
`backend/functions/translation/__tests__/translateChunk.test.ts`, around line 980

The global `beforeEach` calls `resetClients()`, which now clears `_testRateLimiter`. That covers all tests that run *after* the `beforeEach` fires. But the existing rate-limiter test at line 980 calls `setRateLimiterForTesting(mockRateLimiter)` and never explicitly tears it down. Relying solely on the next `beforeEach` cycle for cleanup is fine as long as test ordering stays stable, but it is fragile: if the test block ever gains an `afterEach` that bypasses `resetClients`, the injected mock will leak. Adding an explicit `afterEach(() => resetClients())` inside the rate-limiter `describe` block — or a single `afterEach` inside the new REGRESSION block — would make the lifecycle self-contained and obvious.

**2. The regression test's teardown assertion does not actually verify clearing**
`backend/functions/translation/__tests__/translateChunk.test.ts`, line 1063

```ts
resetClients();
expect(mockRateLimiter.acquire).toHaveBeenCalledTimes(1); // no additional calls after reset
```

This asserts that `acquire` was not called *again* after `resetClients()`, but it does not confirm that `_testRateLimiter` is now `null`. A future change could nullify the variable but set it to something else (or clear it incorrectly) and the assertion would still pass. A stronger teardown verification would be:

```ts
resetClients();
// Confirm the injected limiter is gone: a subsequent handler invocation
// must not reach mockRateLimiter.acquire (it would hit the real singleton path instead).
expect(mockRateLimiter.acquire).toHaveBeenCalledTimes(1);
```

The assertion as written is correct in spirit, but the comment overstates what it proves. Consider rewording: `// resetClients() should have cleared the injected mock; no further calls expected` and leave a note that a full end-to-end clear is verified indirectly via the singleton path taken by all other tests.

**3. `setRateLimiterForTesting` placement relative to `resetClients`**
`backend/functions/translation/translateChunk.ts`, lines 47–61

`resetClients` is exported first, then `setRateLimiterForTesting`. In the existing codebase pattern, `resetClients` is the sole named test-escape-hatch. Having the new export immediately after — before the public interfaces — is fine, but the JSDoc on `setRateLimiterForTesting` (line 54) warns against production use but does not cross-reference `resetClients()` as its counterpart teardown. A single sentence — `Call resetClients() in afterEach to clear the injected value.` — would close the loop for any maintainer reading just the JSDoc.

---

### Suggestions

**4. The `handler.length === 1` test is clever but brittle under TypeScript compilation**
`backend/functions/translation/__tests__/translateChunk.test.ts`, line 1014

`Function.prototype.length` counts declared formal parameters in the compiled JS output. TypeScript parameters with default values are excluded from `.length`. If someone were to add a parameter with a default — `handler(event: TranslateChunkEvent, _unused = undefined)` — the test would still pass while re-introducing the bug. The test is a useful first line of defence and the comment explaining `handler.length` semantics is excellent. Just be aware it is not an airtight contract against all future re-introductions; the JSDoc warning on the handler itself (lines 93–98) is the more durable protection.

**5. Naming: `setRateLimiterForTesting` vs. codebase precedent**
The existing escape hatch is `resetClients()` — a verb-noun pair with no `ForTesting` suffix. The new function appends `ForTesting`, which is more explicit. Both conventions are defensible; the minor inconsistency is unlikely to cause confusion given that the function is only imported in one test file. If the codebase grows more test-DI functions, converging on one style (either always suffix `ForTesting` or never) would be worth a follow-up.

---

### Positive

- The root-cause analysis in the PR description is precise and reproducible (bundle coordinate, sha256 reference, confirmed single `.acquire` call site). That writeup deserves to live as a comment in the commit history — and it does, via the PR.
- The JSDoc on the handler (lines 91–98 of `translateChunk.ts`) is exactly the right length: long enough to explain the why, short enough to be read. Future contributors will not need to rediscover this Lambda runtime quirk.
- Clearing `_testRateLimiter` inside `resetClients()` is the correct lifecycle choice. It means no new teardown contract is required — existing tests that call `resetClients()` in `beforeEach` automatically get the clearing behaviour.
- The test comment at line 976–979 mirrors the production JSDoc and explains the anti-pattern in context, right where someone would reach for it. That is high-value documentation.

leixiaoyu:

**TL;DR:** Runtime perf delta of the code change itself is **negligible to slightly positive**. The operationally significant performance story is *not* in the diff — it's the step-change in Lambda execution profile that lands the moment this fix deploys. Flagging that below so cost/concurrency dashboards aren't read incorrectly during the post-deploy window.

---

### Critical
None. This is a correctness fix; no perf regressions introduced.

---

### Recommendations

**1. Refresh Lambda concurrency / cost projections post-deploy (operational, not code).**
Since `2026-03-19`, every `lfmt-translate-chunk-LfmtPocDev` invocation has failed in <100 ms with `TypeError: i.acquire is not a function` before reaching `loadJob`, `loadChunk`, the rate limiter, or the Gemini API. After this PR deploys, invocations will run the full path:

| Phase | Pre-fix duration | Post-fix expected |
|---|---|---|
| Cold start | ~400-800 ms (unchanged: same imports, same module-init env reads) | ~400-800 ms |
| Warm invocation | <100 ms (TypeError) | ~3-15 s typical (DDB GetItem + S3 GetObject + `rateLimiter.acquire` DDB conditional write + Gemini 2.5 Flash call + S3 PutObject + DDB UpdateItem). Up to 120 s on Gemini retries. |
| Billed GB-seconds per invocation @ 512 MB | ~0.05 GBs | ~1.5-7.5 GBs (30-150x) |

The Map state runs `maxConcurrency: 10`. Document of N chunks now bills ~N successful invocations instead of N failures. Pre-merge dashboards anchored to the broken state will look like a "regression" post-deploy — they aren't. Worth a note on the deploy PR / runbook.

**2. Expect a step-change drop in CloudWatch Logs ingestion volume.**
Pre-fix: every invocation logs `Starting chunk translation` (info) + the `TypeError` stack (error) + `Failed to update job status` chains. Post-fix: same `Starting chunk translation` + the normal success path (Estimated token usage / Translation completed / Job progress updated). Net log lines per invocation are roughly comparable, but error-tier logs collapse to near zero. If anyone has a CloudWatch Logs $ alert keyed on error-rate, expect it to clear; if keyed on total ingestion bytes, no meaningful change.

---

### Suggestions

**3. `setRateLimiterForTesting` in the production bundle — verified non-issue.**
The bundling config in `lfmt-infrastructure-stack.ts:1027-1032` uses `minify: true` with esbuild defaults. Because `setRateLimiterForTesting` is a top-level `export function` declaration with no side effects and is never imported by any production caller (verified: only `__tests__/translateChunk.test.ts` references it), esbuild's DCE eliminates it from the entry bundle. No `/*#__PURE__*/` annotation needed — that hint is for *call expressions*, not function declarations. Leave as-is.

**4. Cold-start timing is unchanged — confirmed.**
The pre-fix bundle snippet `(ac||(ac=new Kl({...})),i=ac)` confirms the singleton was already lazy-initialized inside the handler body (not at module load). The new source preserves this exact pattern (`if (!distributedRateLimiter)` at `translateChunk.ts:127`). Module-level work (`new DynamoDBClient({})`, `new S3Client({})`, four `getRequiredEnv()` calls at lines 29-35) is identical pre/post. Cold-start delta = **zero**.

**5. Per-invocation overhead delta — nominal.**
Pre-fix, on the very first invocation per container, the OR-pattern *did* construct one `DistributedRateLimiter` (then crashed at `i.acquire`). On every subsequent warm invocation in that container, `ac` was already set, so the OR short-circuited and no new object was constructed. So the "constructed a `new Kl()` on every invocation" framing in the prompt is not quite right — it was once per container, same as now. The fix preserves singleton reuse; warm-path object-construction cost delta = **zero**.

---

### Positive

- **Lazy-init pattern preserved.** `distributedRateLimiter` and `geminiClient` both stay container-singletons (`translateChunk.ts:113, 127`). Good for warm-path latency.
- **No new imports, no new module-level work, no new I/O on the hot path.** `_testRateLimiter` is a single null-initialized module variable; `setRateLimiterForTesting` is one assignment. Zero measurable cost.
- **Fast-fail on rate-limit errors retained** (`retryable: true`, ~0 latency cost beyond the DDB conditional write). Step Functions retry semantics unaffected.
- **No change to `maxConcurrency: 10` Map state** — the team's parallel-translation throughput design is now actually exercisable for the first time since 2026-03-19.

---

### Bottom line
Code-level perf impact: **negligible** (and I mean it — there is no measurable warm-path delta). Operational impact: **large and positive but visible as a step-change in dashboards** — make sure whoever watches the post-deploy graphs knows the prior 39-day baseline was a 100% failure floor, not a healthy state.

leixiaoyu:

### Critical

**`handler.length === 1` does not guard against default-value parameters.**
The first regression test (`handler.length === 1`) has a well-known blind spot: JavaScript/TypeScript does not count parameters that have default values in `Function.prototype.length`. If a future developer writes `handler = async (event: TranslateChunkEvent, _ctx = undefined)`, `handler.length` remains `1` and the test passes — while the bug is reintroduced. The guard is better than nothing, but it cannot be the sole structural check.

**Recommended addition** — `backend/functions/translation/__tests__/translateChunk.test.ts`, inside the existing `REGRESSION #150` describe block:

```ts
it('handler must not receive a Lambda context object as the rate limiter', async () => {
  // Simulate exactly what AWS Lambda runtime does: handler(event, context)
  const lambdaContext = {
    functionName: 'lfmt-translate-chunk',
    awsRequestId: 'test-request-id',
    getRemainingTimeInMillis: () => 30000,
    // intentionally no .acquire() method
  };

  // Arrange standard mocks so the handler can reach the rate-limiter branch
  dynamoMock.on(GetItemCommand).resolves({ Item: createMockJob({ ... }) } as any);
  s3Mock.on(GetObjectCommand).resolves({ Body: createMockStream(JSON.stringify({ primaryContent: 'x', chunkId: 'c0' })) } as any);

  const event: TranslateChunkEvent = { jobId: 'job-ctx-test', userId: 'user-123', chunkIndex: 0, targetLanguage: 'es' };

  // Call with two arguments exactly as Lambda runtime does
  // @ts-expect-error — intentionally testing runtime call shape
  await expect(handler(event, lambdaContext)).resolves.not.toThrow();
  // If context leaked into rate-limiter slot, context.acquire() would throw TypeError
});
```

This test directly reproduces the original failure mode regardless of how the handler signature is written.

---

### Recommendations

**`resetClients()` teardown is not directly proven to clear the injection.**
The second regression test calls `resetClients()` and then asserts `mockRateLimiter.acquire` was called exactly once — but this only proves no *further* calls occurred after reset, not that `_testRateLimiter` was set back to `null`. A subsequent `handler(event)` invocation after the reset (using a fresh DynamoDB/S3 mock setup) that completes *without* calling `mockRateLimiter.acquire` would make the assertion airtight.

**Recommended addition** — inside the same `REGRESSION #150` describe block, after the existing second test:

```ts
it('resetClients() must set _testRateLimiter to null so the next invocation uses the real singleton path', async () => {
  const sentinelLimiter = { acquire: jest.fn() } as any;
  setRateLimiterForTesting(sentinelLimiter);
  resetClients();

  // After reset, a handler invocation must NOT reach sentinelLimiter.acquire.
  // The handler will fail at geminiClient initialization (no secret), which is
  // fine — the point is acquire was never called.
  dynamoMock.on(GetItemCommand).resolves({ Item: createMockJob({ jobId: 'job-post-reset', ... }) } as any);
  s3Mock.on(GetObjectCommand).resolves({ Body: createMockStream(JSON.stringify({ primaryContent: 'y', chunkId: 'c0' })) } as any);
  secretsMock.on(GetSecretValueCommand).rejects(new Error('no secret')); // force early failure

  const result = await handler({ jobId: 'job-post-reset', userId: 'user-123', chunkIndex: 0, targetLanguage: 'es' });
  expect(result.success).toBe(false); // failed on secret, not on acquire
  expect(sentinelLimiter.acquire).not.toHaveBeenCalled(); // key assertion
});
```

**`setRateLimiterForTesting` export is only tested transitively.**
There are no dedicated unit tests for the export itself (e.g., that it accepts a `null`-like value, or that calling it twice replaces the first mock). The current test coverage is adequate for the fix's stated goal, but a direct smoke test of the setter contract would be a cheap addition.

---

### Suggestions

- The `beforeEach` at line 89 already calls `resetClients()`, which will clear `_testRateLimiter` between all tests. This is correct and means the test-state isolation is sound — no leakage risk from the DI injection in the rate-limit tests to other tests in the suite.
- The integration test file (`translation-flow.integration.test.ts`) exercises the Lambda end-to-end through the real AWS API Gateway, but it does not invoke the handler function directly, so it provides no unit-level guard against signature regressions. That is expected — integration tests are not the right layer for this check. The unit layer is the correct home for the regression tests added in this PR.

---

### Positive

- The naming of both regression tests is precise and self-documenting; the inline comment block explaining the root cause (Lambda context passed as 2nd arg, `acquire is not a function`) is exactly the level of documentation a future reviewer needs.
- Removing `_rateLimiter` from the handler signature and routing DI through `setRateLimiterForTesting()` is the correct structural fix — it eliminates the failure mode at the call site rather than working around it.
- All 31 existing tests continuing to pass confirms no behavioral regression.
- The `beforeEach` call to `resetClients()` (line 89) ensures `_testRateLimiter` is always null at test start, so the new DI path cannot leak across test cases — good hygiene.
- The PR's guard-rail (JSDoc comment on the handler function explaining *why* only one parameter) is a strong deterrent against the regression being reintroduced silently during future refactors.

**Bottom line:** The fix itself is correct and safe. The two regression tests added are valuable but have one exploitable gap — they do not exercise the actual Lambda runtime call shape `(event, context)`. Adding a single test that calls `handler(event, lambdaContext)` with a real Lambda-context-shaped object (no `.acquire`) would give the regression suite a gap-free guarantee.

leixiaoyu:

**Architectural impact: Low.** Surgical fix to a single Lambda; no boundary, dependency, or contract changes beyond removing one parameter and adding one named test seam. The chosen pattern is consistent with what already exists in this file. The bug itself, however, exposes one missing convention worth fixing once across the codebase.

### Pattern compliance checklist

- [x] Test-DI pattern matches the file's existing convention (`resetClients()` at `translateChunk.ts:47`; `setRateLimiterForTesting()` at `translateChunk.ts:59` is the same module-level-mutable-singleton style).
- [x] Symmetric teardown: `resetClients()` clears `_testRateLimiter` (`translateChunk.ts:50`) — no need for callers to remember two cleanup calls. No footgun.
- [x] Singleton lifecycle preserved: `geminiClient` and `distributedRateLimiter` remain lazy-init across warm-container invocations (`translateChunk.ts:37-39`, `113-137`). The new `_testRateLimiter` short-circuits the singleton without mutating it, so warm-container behavior is unaffected.
- [x] No other handler in `backend/functions/` carries the same latent bug. Verified all 10 handlers (`auth/login.ts`, `auth/register.ts`, `auth/refreshToken.ts`, `auth/resetPassword.ts`, `auth/getCurrentUser.ts`, `jobs/uploadRequest.ts`, `jobs/uploadComplete.ts`, `jobs/startTranslation.ts`, `jobs/getTranslationStatus.ts`, plus this one) — all use a single-event signature.

### Critical
None.

### Recommendations

**1. Adopt the `Handler` type from `@types/aws-lambda` to make this class of bug impossible at the type level.**
`@types/aws-lambda@^8.10.138` is already a declared dependency (`backend/functions/package.json:43`) but is currently not used to type any handler signature in `backend/functions/`. Replacing the bare arrow signature with the SDK-provided `Handler<E, R>` alias would have caused TypeScript to reject the previous `(event, _rateLimiter?)` signature outright, because `Handler` fixes the second parameter as `Context`. Suggested change at `translateChunk.ts:102`:

```ts
import type { Handler } from 'aws-lambda';
export const handler: Handler<TranslateChunkEvent, TranslateChunkResponse> =
  async (event) => { ... };
```

This is the architectural fix — it pushes the Lambda runtime contract into the type system, so a future developer cannot reintroduce a misuse of the second positional argument without TypeScript flagging it. The current PR catches the issue with a runtime `handler.length === 1` regression test (`translateChunk.test.ts:1014`), which is good defense-in-depth, but a compiler-enforced contract is strictly stronger and applies uniformly without per-handler boilerplate. Worth filing as a follow-up to apply across all 10 handlers.

### Suggestions

**2. The "cleared by `resetClients()`" assertion is a tautology and doesn't actually verify what its name promises.**
At `translateChunk.test.ts:1062-1063`, the test calls `resetClients()` and then asserts `mockRateLimiter.acquire` was called exactly once — but the handler is never re-invoked after the reset, so of course the call count is unchanged. The test name says "must be cleared by resetClients()" but the assertion only proves "no time travel happened." To actually verify the contract, either (a) re-invoke `handler(event)` after `resetClients()` and assert the mock is NOT called (because the real singleton would now be used), or (b) assert directly that calling the handler after reset re-initializes `distributedRateLimiter` from scratch. As written, this test would pass even if `resetClients()` didn't clear `_testRateLimiter` at all.

**3. Consider naming the test seam to match the codebase's existing "for testing" idiom.**
`resetClients()` uses a comment-based "(for testing)" annotation (`translateChunk.ts:45`), while the new function uses a `ForTesting` suffix in its name. Both work; picking one convention and applying it consistently makes intent obvious at every call site without reading JSDoc. Minor — but if more test seams get added later, a single convention scales better.

### Positive

- The JSDoc on `handler` (`translateChunk.ts:91-101`) explicitly documents *why* the signature is one-arg and references the issue number — exactly the kind of architectural commentary that prevents repeat regressions.
- Adding `handler.length === 1` as a regression test (`translateChunk.test.ts:1011-1015`) is a clever, low-cost runtime guard that doesn't require type-checking infrastructure to run.
- The fix is correctly scoped: it does not over-engineer (no DI container, no factory, no class refactor), which would have been disproportionate to a one-file bug.

### Long-term implications

- **Positive**: codifies the rule "Lambda handlers take one argument" in this file via tests + JSDoc.
- **Neutral**: keeps the module-level-mutable-singleton test pattern. This pattern doesn't compose well if multiple test files import the same module in parallel (Jest `--workerIdleMemoryLimit` and isolated module registries usually save you, but it's a known sharp edge), but that risk already existed with `resetClients()`. No regression.
- **Latent**: until Recommendation 1 is applied across all handlers, the next person who adds a Lambda is one IDE autocomplete away from re-introducing the same shape of bug in a different file. The runtime regression test only protects `translateChunk`.

leixiaoyu:

5 specialist agents reviewed PR #167 in parallel. Individual reports:

- Code Quality — [#issuecomment-4353700836](https://github.com/leixiaoyu/lfmt-poc/pull/167#issuecomment-4353700836)
- Security — [#issuecomment-4353700689](https://github.com/leixiaoyu/lfmt-poc/pull/167#issuecomment-4353700689)
- Architecture — [#issuecomment-4353703142](https://github.com/leixiaoyu/lfmt-poc/pull/167#issuecomment-4353703142)
- Performance — [#issuecomment-4353702918](https://github.com/leixiaoyu/lfmt-poc/pull/167#issuecomment-4353702918)
- Test Coverage — [#issuecomment-4353703075](https://github.com/leixiaoyu/lfmt-poc/pull/167#issuecomment-4353703075)

---

### Verdict: Approve **after addressing the regression-test gap**

The fix itself is correct and approved by Security, Performance, Architecture, and Code Quality (no Critical from any of them). Test Coverage flagged **one Critical regression-test gap** that's worth fixing in this PR before merge — it's directly tied to the bug being fixed.

---

1. **`handler.length === 1` cannot catch a reintroduced default-value 2nd parameter.** _Flagged by test-automator (Critical); also raised as a Suggestion by code-reviewer._ A future contributor adding `handler(event, _ctx = undefined)` would have `handler.length === 1` (default-value params don't count) and the regression test would silently pass — re-opening exactly the bug this PR is fixing. The strongest guard is an integration-style test that calls `handler(mockEvent, mockLambdaContext)` with a Lambda-context-shaped object that has no `.acquire()` method, directly reproducing the original failure mode. Add to `translateChunk.test.ts` in the same `REGRESSION #150` describe block.

---

### Recommendations (should address — repeated across reviewers)

2. **Tautological teardown assertion in the regression test.** _Flagged by 3 reviewers: code-reviewer (R), architect (S), test-automator (R)._ The current proof that `resetClients()` clears the injected mock asserts `toHaveBeenCalledTimes(1)` AFTER the reset — but the handler is never re-invoked after the reset, so the assertion is trivially true regardless of whether the injection was actually cleared. Fix: after `resetClients()`, invoke `handler(event)` again and assert the mock was NOT called (the singleton must have been re-initialized fresh).

3. **Adopt `Handler<E, R>` from `@types/aws-lambda`** as the type for the exported `handler`. _Flagged by architect-reviewer (R)._ The type is **already declared in `backend/functions/package.json:43`** but unused for handler signatures across all 10 handlers in this codebase. Using it would have rejected the original 2-arg signature at compile time, replacing the runtime `handler.length === 1` check with a stronger compiler-enforced contract. High value for permanent prevention; small change in scope. Consider as an optional follow-up if it expands this PR's footprint too much, but flag it for a near-term cleanup.

4. **Existing rate-limiter test block doesn't explicitly tear down `_testRateLimiter`** — relies entirely on `beforeEach` ordering. _Code-reviewer (R)._ Add an explicit `afterEach` (or move the cleanup to inside the test's `try/finally`) so the test's intent is self-documenting and not load-bearing on test framework execution order.

---

### Suggestions (nice-to-have)

- **Defense-in-depth: throw inside `setRateLimiterForTesting()` when `AWS_EXECUTION_ENV` is set** (security). The export is now public API — a future mistaken caller from production code would silently disable distributed rate limiting and burn Gemini quotas. Cheap guard.
- **ESLint `no-restricted-imports` rule scoping the symbol to `__tests__/`** (security). Even cheaper guard if you don't want runtime overhead.
- **Naming inconsistency: `setRateLimiterForTesting` vs `resetClients`** (code, architect). Two competing test-seam conventions in the same file. Pick one.
- **`setRateLimiterForTesting()` only tested transitively** (test-automator). Adding a small direct contract test (set → assert get → reset → assert cleared) would document the API.

---

### Operational notes (post-deploy, not code changes)

- **Refresh Lambda cost/concurrency projections** (performance). Invocations go from <100ms TypeError failures to 3-15s real work — **30-150× billed GB-seconds per invocation at 512 MB**. Existing capacity planning was anchored to the broken state.
- **Expect a step-change drop in error-tier CloudWatch log volume** (performance). Positive impact on log costs.
- **Parallel translation (`maxConcurrency: 10`) is now actually exercisable for the first time since 2026-03-19** (performance). Worth validating Gemini rate-limit behavior on the first real run.

---

### Positives (consensus)

- The handler signature change correctly identifies the AWS Lambda runtime contract violation. Inline JSDoc and the in-test anti-pattern comment effectively warn future maintainers.
- Module-level `_testRateLimiter` state is safely cleared by `resetClients()` in `beforeEach` and unreachable from production code paths.
- The original TypeError leaked no Lambda context to API consumers (only to CloudWatch).
- **Confirmed no other handler in `backend/functions/` has the same 2nd-param DI anti-pattern** (security + architect both grepped).
- `setRateLimiterForTesting` is dead-code-eliminated by esbuild's tree-shaker (no production importers, `minify: true`).
- Cold-start unchanged; lazy-init preserved; warm-path singleton reuse cost delta is zero.
- `handler.length === 1` is a useful structural invariant **for the no-default-param case**.

---

### Recommendation

**Fix Critical #1 (the integration-style regression test) and Recommendation #2 (the tautological teardown) in this PR.** Both are tiny test additions in the same file and directly strengthen the proof that #150 cannot recur — fitting the spirit of "lock in the fix, not just apply it." Recommendation #3 (`Handler<E,R>` typing) is high-value for permanent prevention but touches all 10 handlers; reasonable to defer to a focused follow-up. Suggestions and operational notes can land in follow-ups.

leixiaoyu:
## OMC Review Follow-up — Critical + Recommendation #2 Applied

Both fixes are test-only, additive on top of `250f1f6`. No production code touched.

### What changed

**`backend/functions/translation/__tests__/translateChunk.test.ts`**

1. **Critical (test-automator):** New test added to the `REGRESSION #150` describe block — `"handler invoked with a Lambda context object must not treat it as a rate limiter"` (lines ~1090–1190). Builds a Lambda-context-shaped mock object that deliberately has **no `.acquire()` method** and invokes `handler(event, mockLambdaContext)` exactly the way AWS Lambda does. Asserts that no `TypeError: ... acquire is not a function` is thrown and that the injected (real DI path) mock limiter was actually called — proving the handler ignores the second argument entirely. This catches the exact regression vector that `handler.length === 1` cannot, since default-value parameters do not count toward `Function.prototype.length`.

2. **Recommendation #2 (code-reviewer + architect-reviewer + test-automator):** Strengthened the existing test `"setRateLimiterForTesting() must be the only DI path and must be cleared by resetClients()"` (lines ~1017–1090). The previous teardown proof asserted `toHaveBeenCalledTimes(1)` immediately after `resetClients()` but never re-invoked the handler — tautologically true regardless of whether the injection was cleared. Now snapshots the mock call count before reset, re-stubs DynamoDB/S3, invokes the handler a second time, and asserts the mock's call count is unchanged (proving the singleton was re-initialized fresh).

### Why

Both items are direct asks from the [OMC consolidated summary](https://github.com/leixiaoyu/lfmt-poc/pull/167#issuecomment-4353713205) — Critical #1 and Recommendation #2.

### Verification

- `npm test -- --testPathPattern=translateChunk`: **32/32 passing** (was 31)
- `npm run lint`: **0 errors** (warnings only — pre-existing `no-explicit-any` pattern, consistent with the rest of the file)
- `npm run format:check`: **clean** (all matched files use Prettier code style)

### Test count delta

`31 → 32` (Fix #1 adds one test; Fix #2 modifies an existing test body, no count change)

### Deliberately deferred

- **Recommendation #3 (`Handler<E, R>` typing across all 10 handlers)** — High-value compile-time guard, but it touches all handlers in `backend/functions/`. Better as a focused follow-up PR than expanding this one's footprint.
- **Suggestions** (`AWS_EXECUTION_ENV` runtime guard inside `setRateLimiterForTesting`, `no-restricted-imports` ESLint rule, `setRateLimiter`/`resetClients` naming consistency, direct contract test for `setRateLimiterForTesting`) — All low-priority defense-in-depth; can roll into the Rec #3 follow-up or stay as separate issues.
- **Recommendation #4 (explicit `afterEach` for `_testRateLimiter` cleanup)** — Marginal benefit on top of the strengthened teardown test now in place; `beforeEach` ordering already covers it.

xlei-raymond (approved):
Detailed Review Findings

  1. Technical Accuracy (The "Lambda Context" Footgun)
  The PR correctly diagnoses why the lfmt-translate-chunk Lambda was throwing TypeError: i.acquire is not a function.
   * The Bug: The handler signature was handler(event, _rateLimiter?). AWS Lambda always passes (event, context) to handlers. Consequently, the context object was being assigned to _rateLimiter. Since context is truthy,
     the code incorrectly assumed a test mock had been provided and attempted to call context.acquire(), resulting in a crash.
   * The Fix: Removing _rateLimiter from the signature and moving to a dedicated setRateLimiterForTesting setter is the correct architectural choice. It strictly decouples the production runtime from the test-injection
     path.

  2. Robustness of Regression Guards
  The team has implemented exceptionally high-quality regression tests, specifically addressing several edge cases from the OMC audit:
   * Integration-Style Guard: The new test handler invoked with a Lambda context object must not treat it as a rate limiter is a brilliant addition. It directly reproduces the original failure by calling handler(event,
     context) with a non-mock object, proving the second argument is now safely ignored.
   * Default-Param Resistance: By using the integration-style test instead of relying solely on handler.length === 1, the fix is now resistant to future contributors who might mistakenly add a second parameter with a
     default value (which doesn't affect .length).
   * Teardown Integrity: The strengthened teardown test now correctly proves that resetClients() clears the injection by re-invoking the handler and asserting the mock is no longer used.

  3. Code Standards & Hygiene
   * Lazy Initialization: The lazy-initialization pattern for singletons is preserved, ensuring no regression in cold-start performance.
   * JSDoc Warnings: The inclusion of a loud "IMPORTANT" warning in the handler JSDoc effectively warns future maintainers against re-introducing the DI parameter.

  4. Verification Results
  I have independently verified the implementation within the worktree:
   * Tests: All 32 tests in translateChunk.test.ts passed, including the 3 new regression guards.
   * Logic: Confirmed the toWireJob and toWireStatus logic remains consistent with the shared-types used by the polling endpoint.

  Verdict: APPROVE (GO)
  This PR is a textbook example of how to fix a production bug with "Honest Engineering." It doesn't just fix the call site; it creates a robust set of automated guards that make the failure mode virtually impossible to
  re-introduce.

  ---

  PR Review for GitHub

  Status: ✅ Approved

  This PR correctly identifies and fixes a critical "handler signature" bug (Issue #150) that was causing 100% of production translations to fail with a TypeError. The implementation is disciplined and demonstrates an
  excellent response to the prior OMC audit feedback.

   1. Resolved Runtime Conflict: The root cause—assigning the mandatory AWS Lambda context object to an optional _rateLimiter parameter—is correctly identified and eliminated. The transition to the
      setRateLimiterForTesting setter provides a clean, side-effect-free path for unit test dependency injection.
   2. Robust Integration-Style Guard: I highly commend the addition of the "Lambda context" regression test. By directly invoking the handler with a mock context object that lacks an .acquire() method, the team has
      created a deterministic guard against the original failure mode that is resistant to future signature changes.
   3. Strengthened Teardown Proof: The updated teardown test correctly snapshots mock call counts and re-invokes the handler post-reset. This provides actual proof that singleton state is not leaking across test
      boundaries, resolving the "tautological assertion" concern from the audit.
   4. Zero-Impact Production Path: The fix maintains the existing lazy-initialization pattern for the DistributedRateLimiter, ensuring zero regression in Lambda cold-start latency or warm-path execution cost.

  Final Verification Results
   * Tests: 32 passed (100% success rate in translateChunk.test.ts).
   * Regression Guard: 100% (Integration-style context call verified).
   * Hygiene: PASS (JSDoc warnings added, dead code eliminated).

  ---
  Status: 100% Production Bug Resolved; Regression Guards Validated.

PR #177, opened 2026-05-02
fix(ci): unblock post-deploy smoke + E2E + production-smoke verification jobs

leixiaoyu:
## Summary

Two pre-existing failure patterns have been blocking every deploy's post-deploy verification gates. Both fixes are minimal and surgical.

### Fix 1 — `__dirname is not defined in ES module scope`

**File**: `frontend/e2e/tests/smoke/production-smoke.spec.ts:30`

Bare `__dirname` is not defined in ES module scope. Playwright loads this spec as ESM, so the file fails to load with:

```
ReferenceError: __dirname is not defined in ES module scope
```

…before a single assertion runs. This crashes **both** `Run Production Smoke Tests` **and** `Run E2E Tests` (both jobs include this spec file).

Applied the standard ESM workaround: import `fileURLToPath` from `node:url` and derive `__dirname` from `import.meta.url`. The existing `SMOKE_FIXTURE_PATH` line is unchanged — it now resolves correctly because `__dirname` is defined.

### Fix 2 — `❌ TEST_PASSWORD environment variable is required`

**File**: `.github/workflows/deploy.yml` (the `Run smoke tests against deployed API` step, ~line 557)

The step read `TEST_PASSWORD` from `secrets.SMOKE_TEST_PASSWORD`, which was never configured in repo settings. The backend smoke test asserts the env var at startup (`backend/tests/smoke/smoke.test.ts:42-46`) and exits 1 with the error above before running anything.

Per the smoke test's own docstring it **registers fresh users each run**, so it just needs a Cognito-policy-compliant password — there's no need for a stable shared secret. The step now generates a per-run password via `openssl rand`, masks it with `::add-mask::`, and exports it inline. This removes the operational dependency on a missing secret entirely and keeps the password ephemeral.

`API_URL` and `USER_POOL_ID` env vars are preserved; only `TEST_PASSWORD: ${{ secrets.SMOKE_TEST_PASSWORD }}` was removed from the env block.

## Verification (local)

- [x] `npx tsc --noEmit -p frontend/tsconfig.json` — clean (no errors mentioning `production-smoke`)
- [x] `npx prettier --check frontend/e2e/tests/smoke/production-smoke.spec.ts` — clean
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/deploy.yml'))"` — YAML parses

## Out of scope / follow-up

After Fix 1 unblocks them, frontend `Run Production Smoke Tests` and `Run E2E Tests` may surface a **new** failure pattern: they read `SMOKE_TEST_EMAIL` and `SMOKE_TEST_PASSWORD` to log in to a known account, and those secrets are also unset in repo settings. That is a larger refactor (it requires either a seeded test account or refactoring the frontend smoke flow to register a fresh user the way the backend smoke test does) and is intentionally **not** addressed here. Tracking as a follow-up.

## Notes

This is a clean re-do of an earlier attempt that was lost when a previous working tree was reset by a colliding agent edit. This PR was prepared in an isolated git worktree on a fresh branch off main, so the diff contains only the two intended files and nothing else.

## Test plan

- [ ] Verify `Run Tests` and `Build Infrastructure` (the two required checks) pass on this PR
- [ ] After merge, monitor the next deploy run and confirm `Run smoke tests against deployed API`, `Run Production Smoke Tests`, and `Run E2E Tests` no longer fail with the two patterns above (they may surface the documented follow-up failure for the frontend smoke jobs — that's expected)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -556,11 +556,24 @@ jobs:
 
       - name: Run smoke tests against deployed API
         working-directory: backend/tests/smoke
-        run: npm run test:smoke
         env:
           API_URL: ${{ needs.deploy-dev.outputs.api_url }}
-          TEST_PASSWORD: ${{ secrets.SMOKE_TEST_PASSWORD }}
           USER_POOL_ID: ${{ secrets.DEV_USER_POOL_ID }}
+        # The backend smoke test (backend/tests/smoke/smoke.test.ts) registers
+        # FRESH users each run (per its docstring) and just needs a password
+        # strong enough to satisfy Cognito's policy. We previously read this
+        # from `secrets.SMOKE_TEST_PASSWORD`, but that secret was never
+        # configured in repo settings — every post-deploy run failed with
+        # `❌ TEST_PASSWORD environment variable is required` before the suite
+        # even started. Generating per-run with openssl removes the operational
+        # dependency entirely and keeps the password ephemeral. `::add-mask::`
+        # ensures the value is redacted from the workflow log if anything
+        # downstream prints it.
+        run: |
+          GENERATED_PASS="$(openssl rand -base64 24 | tr -d '/+' | head -c 20)Aa1!"
+          echo "::add-mask::$GENERATED_PASS"
+          export TEST_PASSWORD="$GENERATED_PASS"
+          npm run test:smoke
 
       - name: Smoke Test Summary
         if: always()
--- a/frontend/e2e/tests/smoke/production-smoke.spec.ts
+++ b/frontend/e2e/tests/smoke/production-smoke.spec.ts
@@ -17,12 +17,20 @@
  */
 
 import * as path from 'path';
+import { fileURLToPath } from 'node:url';
 import { test, expect } from '@playwright/test';
 
 // Smoke tests use a longer timeout because they run against production.
 // 3 minutes is sufficient when we use the ~1 KB smoke-test-minimal.txt
 // fixture (one chunk, deterministic translation time).
 test.setTimeout(180000); // 3 minutes
+
+// __dirname is not defined in ES module scope (Node ESM). Compute it from
+// import.meta.url instead. Without this, the test file fails to load with
+// `ReferenceError: __dirname is not defined in ES module scope` and every
+// post-deploy verification job that imports it (Production Smoke Tests,
+// E2E Tests) crashes before running a single assertion.
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 // Path to the tiny fixture we upload during the smoke test. Keeping this
 // ~1 KB guarantees the test completes inside the timeout regardless of


leixiaoyu:

**Verdict: Pass — no perf impact.**

CI-infra only; zero production code touched.

**Runtime delta:**
- `openssl rand -base64 24 | tr | head` + `::add-mask::` + `export`: ~5–15 ms total. Noise on a GitHub runner.
- ESM `__dirname` shim (`fileURLToPath(import.meta.url)`): one-time module-load cost, sub-millisecond. No per-test overhead.

**Wall-clock recovery (the real win):**
Pre-PR, both Production Smoke Tests and post-deploy E2E crashed at import time / before suite start, so deploy 'completed' in seconds-of-failure. Post-PR, those jobs will actually execute — expect deploy wall-clock to grow by the suite's natural runtime (smoke timeout budget is 180s/test; full suite likely a few minutes). That's restored signal, not regression.

No follow-ups recommended.

leixiaoyu:

**Verdict: Pass — clean, well-commented.**

### ESM `__dirname` fix (`production-smoke.spec.ts`)

Standard pattern, correct form. Comment explains the ReferenceError root cause and which CI jobs it broke. No issues.

### Workflow-generated TEST_PASSWORD (`deploy.yml`)

- Password construction: `openssl rand -base64 24 | tr -d '/+' | head -c 20` yields 20 alphanumeric chars; appending `Aa1!` guarantees all four Cognito character classes (upper, lower, digit, symbol). Policy satisfied.
- `::add-mask::` is set before `export TEST_PASSWORD` and before `npm run test:smoke` — masking is active before any downstream logging can occur. Correct placement.
- Comment explains the operational gap (secret never configured, every post-deploy run failed before a single assertion) and the ephemeral design intent. Explains WHY, not just WHAT.

No concerns. Both fixes are minimal, targeted, and safe to merge.

leixiaoyu:

**Verdict: Pass — testing rationale sound.**

**Fix #1 (`production-smoke.spec.ts` ESM `__dirname`):** Correct and necessary. `package.json` declares `"type": "module"`, so `__dirname` is undefined at module scope. The `fileURLToPath(import.meta.url)` pattern is the standard ESM equivalent. `SMOKE_FIXTURE_PATH` resolves correctly post-fix. No unit test needed for this path computation — it is trivially verified by the spec loading without a `ReferenceError`.

**Fix #2 (`deploy.yml` generated `TEST_PASSWORD`):** The backend smoke test registers fresh users per run and only needs a Cognito-policy-compliant password. Generating it with `openssl rand` + `::add-mask::` is correct and removes a misconfigured (never-set) secret dependency. Cannot be tested with `act` locally without AWS credentials, but `shellcheck` would pass — the shell logic is straightforward. The password format (`base64` stripped of `/+`, truncated to 20 chars, suffixed `Aa1!`) satisfies typical Cognito complexity rules.

Neither fix removes or weakens existing test coverage. Both are pure unblocking repairs to test infrastructure.

leixiaoyu:

**Impact: Medium** — both fixes are correct in isolation, but expose a workflow-level inconsistency in how "smoke test credentials" are modeled.

### Findings

**1. `openssl rand` vs pre-shared secret — wrong shape applied uniformly is the risk, not this fix**
This fix is correct for `backend/tests/smoke` (docstring confirms it registers fresh users — credentials are *ephemeral*). Do **not** copy this pattern to the frontend smoke step at `deploy.yml:499-507`: `production-smoke.spec.ts:36` reads `SMOKE_TEST_PASSWORD` to log into a *known* account (and throws if unset, line 46). Two distinct shapes are needed:
- backend smoke = ephemeral → generate per-run (this PR, correct)
- frontend smoke = known account → real secret, OR refactor the spec to register-fresh like the backend does

Recommend documenting this distinction in a comment block at both call sites so future contributors don't \"unify\" them incorrectly.

**2. Other workflow steps with the same broken pattern: YES**
`.github/workflows/deploy.yml:505-506` still references `secrets.SMOKE_TEST_EMAIL` / `secrets.SMOKE_TEST_PASSWORD` — both unset per PR body. After fix #1 unblocks the spec from loading, this step will fail at line 46 of the spec (`throw new Error(...)`). This is the \"NEW failure pattern\" the PR predicts. Out of scope for *this* PR, but should be filed as a follow-up issue immediately so the prediction doesn't get lost. Two viable architectures:
  - (a) Set the secrets + provision the account via CDK (couples CI to a long-lived Cognito user — operational burden, drift risk)
  - (b) Refactor frontend smoke to self-provision like backend (consistent shape across both smoke suites, zero shared state — preferred)

Strong recommendation: **(b)**. It eliminates the \"secret never set\" failure mode permanently and gives both smoke suites the same architectural shape.

**3. ESM `__dirname` — fix-as-discovered is fine, but codify the pattern**
Grep shows 8 specs in `frontend/e2e/tests/translation/` already use the `fileURLToPath(import.meta.url)` pattern (consistently, with the same `// ES module equivalent of __dirname` comment). `production-smoke.spec.ts` was the only outlier. No preemptive sweep needed — the convention exists, this PR just brings the smoke spec into line with it.

Minor nit: the existing 8 specs use `import { dirname } from 'path'` + `fileURLToPath`; this PR uses `path.dirname(...)`. Functionally identical, but inconsistent with the established idiom. Consider matching the existing pattern for grep-ability.

### Pattern Compliance Checklist
- [x] No new abstractions introduced (correct — these are tactical fixes)
- [x] Backend fix matches the test's actual semantics (ephemeral users)
- [ ] Frontend smoke credentials shape mismatch — pre-existing, not introduced here, but adjacent
- [~] ESM `__dirname` idiom — works, slight stylistic drift from sibling specs

### Long-Term Implications
The real architectural smell exposed here is that **\"smoke test\" means two different things in this repo** (self-provisioning vs. known-account login) without that distinction being documented anywhere. Either unify the contract (option b above) or document the divergence. Leaving it implicit guarantees this confusion resurfaces.

### Recommendation
Approve as-is for unblocking deploy. **File a follow-up issue immediately** to refactor frontend smoke to self-provision, eliminating `SMOKE_TEST_EMAIL`/`SMOKE_TEST_PASSWORD` secrets entirely and giving both suites a consistent shape.

leixiaoyu:

**Verdict: APPROVE with one MEDIUM finding to fix in a follow-up.**

### Findings

**[MEDIUM] Stale `secrets.SMOKE_TEST_PASSWORD` reference still present**
`.github/workflows/deploy.yml:506` (the **frontend** Playwright `Run production smoke tests` step) still wires `SMOKE_TEST_PASSWORD: ${{ secrets.SMOKE_TEST_PASSWORD }}`. This PR only fixes the backend `smoke-tests` job (line ~562). That job will continue to fail on every post-deploy run with the same root cause. `frontend/e2e/tests/smoke/production-smoke.spec.ts:36-46` still throws `SMOKE_TEST_PASSWORD environment variable is required`. Recommend a follow-up PR applying the same `openssl rand` + `::add-mask::` pattern there, or document why the frontend smoke job is intentionally left broken.

**[LOW / informational] Password entropy & Cognito policy: OK**
- `openssl rand -base64 24` = 192 bits; after `tr -d '/+'` (~25% drop) and `head -c 20`, worst case ~120 bits remain. Far above any brute-force threshold for an ephemeral per-run user.
- `Aa1!` suffix deterministically satisfies Cognito's default policy (upper + lower + digit + symbol, len >= 8). Final length 24.
- Note: the suffix is constant, so the *effective* secret entropy is the 20 random chars, which is still fine.

**[PASS] `::add-mask::` ordering**
Mask is emitted in the same `run:` block before `npm run test:smoke`, so any downstream echo of `$TEST_PASSWORD` would be redacted. No subsequent step in `smoke-tests` references the var. OWASP A09:2021 (Logging Failures) addressed.

**[PASS] No password logging in test code**
`backend/tests/smoke/smoke.test.ts` was audited — no `console.log`/error message includes the password value (only the literal string `TEST_PASSWORD` in the env-var-missing error).

**[PASS] ESM `__dirname` workaround**
Pure Node module-resolution fix. No security surface.

### Recommendation
Merge as-is to unblock the backend smoke job, then open a follow-up to apply the same fix at deploy.yml:506.

leixiaoyu:

5 specialist agents reviewed PR #177. Individual reports:

- Code Quality — [issuecomment-4364540506](https://github.com/leixiaoyu/lfmt-poc/pull/177#issuecomment-4364540506) (Pass)
- Security — [issuecomment-4364541602](https://github.com/leixiaoyu/lfmt-poc/pull/177#issuecomment-4364541602) (1 Medium — known out-of-scope)
- Architecture — [issuecomment-4364541107](https://github.com/leixiaoyu/lfmt-poc/pull/177#issuecomment-4364541107) (Medium impact)
- Performance — [issuecomment-4364540387](https://github.com/leixiaoyu/lfmt-poc/pull/177#issuecomment-4364540387) (Pass)
- Test Coverage — [issuecomment-4364540743](https://github.com/leixiaoyu/lfmt-poc/pull/177#issuecomment-4364540743) (Pass)

---

### Verdict: **APPROVED — ready to merge with documented follow-up**

| Reviewer | Critical | Recommendations | Verdict |
|---|---|---|---|
| Code Quality | 0 | 0 | Pass — safe to merge |
| Security | 0 | 1 Medium (out-of-scope) | Pass for what's in scope |
| Architecture | 0 | 1 (out-of-scope) | Medium impact, well-bounded |
| Performance | 0 | 0 | Pass — recovered signal, not regression |
| Test Coverage | 0 | 0 | Pass |

**No Critical findings.** Both fixes verified correct; the only finding is consensus around the out-of-scope frontend smoke step which PR body already documents.

### What landed in this PR (verified Pass)

- **`__dirname` ESM fix** in `production-smoke.spec.ts` correct. Required because `package.json` sets `"type": "module"`. Pattern matches Node ESM standard.
- **Backend smoke `TEST_PASSWORD` workflow generation** correct shape:
  - 24-char base64 → 20-char + `Aa1!` suffix satisfies all 4 Cognito character classes
  - ~120 bits entropy worst case
  - `::add-mask::` correctly placed before any potential downstream echo
  - No password logging in `smoke.test.ts`
  - Removes operational dependency on a never-configured secret

### Out-of-scope finding (consensus from security + architect — already in PR body)

The **frontend Playwright smoke step** (`deploy.yml:505-506`) still references the same never-set `secrets.SMOKE_TEST_PASSWORD`, AND `production-smoke.spec.ts:36,46` requires a *known* account (login flow, not register). After this PR's `__dirname` fix unblocks the file from loading, that job will surface a NEW failure (login error instead of TypeError).

This was explicitly called out in the PR body as out-of-scope — the right fix is to refactor the frontend smoke to also self-provision (consistent shape with the backend smoke). Reasonable as a focused follow-up.

### Performance highlight

Real change post-merge is **recovered signal**: pre-PR the smoke + E2E jobs crashed before any assertion (deploy "finished" in seconds-of-failure); post-PR they actually run, so deploy wall-clock grows by the suite's natural runtime (per-test budget 180s; full suite a few min). That's recovered coverage, not regression.

### Minor stylistic nit (non-blocking)

This PR uses `path.dirname(...)` while 8 sibling specs in `frontend/e2e/tests/translation/` use `import { dirname } from 'path'` directly. Slight drift — easy to align in a follow-up if anyone cares.

### Recommendation

Ready to merge. The frontend smoke refactor is a natural follow-up issue; can be filed as such if you want it tracked.

xlei-raymond (approved):
PR Review: Approved

  Status: ✅ Approved

  This PR is a critical infrastructure update that correctly identifies and resolves two distinct "fail-on-startup" patterns in our post-deploy verification pipeline. By fixing the ESM compatibility issue in the
  Playwright spec and transitioning the backend smoke tests to ephemeral, workflow-generated credentials, the team has restored essential verification signal that has been masked for some time.

   1. Standard-Compliant ESM Shimming: The use of fileURLToPath(import.meta.url) to derive __dirname is the correct industry pattern for Node.js ESM. This resolves the ReferenceError that previously prevented the
      production smoke specs from even loading.
   2. Operational Pragmatism: Generating high-entropy, Cognito-compliant passwords per-run for the backend smoke tests is an excellent solution. It eliminates the "secret-not-configured" failure mode while respecting the
      test's existing self-provisioning semantics.
   3. Security-First Workflow: The implementation correctly prioritizes the use of ::add-mask:: before exporting the generated password. This ensures that our audit trail remains clean and protected from accidental
      secret leakage in logs.
   4. Transparent Technical Debt: I highly commend the PR's "Honest Engineering" approach—explicitly identifying that the frontend smoke tests will likely surface a new failure pattern (login error) once unblocked. This
      allows us to land these foundational fixes immediately while tracking the refactor in Issue #178.

  Final Verification Results
   * Logic Correctness: 100% (ESM shim and password entropy verified).
   * YAML Integrity: PASS (Syntax checked).
   * Standards Compliance: 100% (Clean lint and documentation).

  ---
  Status: 100% Infrastructure Verified; Verification Pipeline Unblocked.

Issue #64, opened 2025-11-10
P3-ARCH: Plan Refactoring of Monolithic CDK Stack to Nested Stacks


leixiaoyu:
As our project grows, the single 'LfmtInfrastructureStack' is becoming large and slow to deploy. We should plan a future architectural refactoring to break this monolith into smaller, nested stacks (e.g., 'FrontendStack', 'ApiStack', 'DatabaseStack'). This will improve separation of concerns, reduce the blast radius of changes, and speed up deployments. This ticket is for creating the initial 'OpenSpec' proposal for this architectural change.

leixiaoyu:
## Closing — OpenSpec proposal delivered by PR #262

The issue body explicitly scopes this ticket to the **initial OpenSpec proposal**:

> "This ticket is for creating the initial 'OpenSpec' proposal for this architectural change."

That deliverable shipped in PR #262 (merged 2026-05-15):

- **Proposal directory**: [`openspec/changes/refactor-cdk-nested-stacks/`](https://github.com/leixiaoyu/lfmt-poc/tree/main/openspec/changes/refactor-cdk-nested-stacks)
- **Files**: `proposal.md`, `design.md`, `tasks.md`, `specs/infrastructure-stack/spec.md`
- **Validation**: `openspec validate refactor-cdk-nested-stacks --strict` → **valid**
- **OMC R1 self-review**: https://github.com/leixiaoyu/lfmt-poc/pull/262#issuecomment-4456103968

### Proposed stack-split (one line)

DataStack (DDB + S3 + Secrets) · AuthStack (Cognito + PreSignUp) · ApiStack (API GW + 15 Lambdas + 7 IAM roles) · TranslationWorkflowStack (Step Functions + exec role) · FrontendStack (S3 + CloudFront + OAC) · SecurityStack (CSP-report Lambda + Lambda@Edge slot)

### Implementation tracking

The proposal's [`tasks.md`](https://github.com/leixiaoyu/lfmt-poc/blob/main/openspec/changes/refactor-cdk-nested-stacks/tasks.md) IS the implementation tracker — 80 sequenced tasks across 7 phases:

- Phase 0 ✅ — proposal scaffolding (this PR)
- Phase 1-7 — implementation (76 tasks, awaits owner approval)

[...]

PR #208, opened 2026-05-07
fix(smoke): implement GET+DELETE /jobs/{jobId}, fix fileSize floor, fix poll URL

leixiaoyu:
## Summary

Three distinct root causes were causing consistent smoke-test failures across every deploy. All three are fixed in this PR.

- **Bug 1** (`should poll job status` → 403): The test was polling `GET /jobs/{jobId}`, which was never registered in API Gateway. Fixed by pointing the test at `GET /jobs/{jobId}/translation-status`.
- **Bug 2** (`TypeError: Cannot read properties of undefined (reading 'jobId')`): Two uploads used `fileSize` below `MIN_FILE_SIZE = 1000`. Fixed: bumped to 1024.
- **Bug 3** (`GET /jobs/{jobId}` and `DELETE /jobs/{jobId}` never existed): Implemented `getJob.ts` and `deleteJob.ts`, wired into CDK.

## Round 2 — OMC follow-ups (commit 4ccd38e)

All 4 Critical + all 11 Recommended applied. Soft-delete deferred per user R3 decision (Issue #209).

| Item | Status | Notes |
|---|---|---|
| Critical 1: requestId threading | Done | Both handlers now extract and pass `requestId` |
| Critical 2: Dedicated deleteJobRole | Done | Role 5 (DeleteJobLambdaRole) isolated from translationRole; infra test asserts translationRole has NO DeleteItem |
| Critical 3: BOLA cross-ownership tests | Done | 2 tests per handler (404 + privacy assertion) |
| Critical 4: Smoke beforeAll guards | Done | `expect(uploadResponse.status).toBe(200)` before each `.data.data.jobId` access |
| Rec 5: ConditionalCheckFailedException → 404 | Done | Single-round-trip delete; CCF → 404 |
| Rec 6: Extract loadJobForUser to jobRepository | Done | `shared/jobRepository.ts` + 5 unit tests; getJob uses it; getTranslationStatus retains ConsistentRead inline (documented) |
| Rec 7: S3 cascade delete | Done | deleteJob deletes uploads/ + documents/ keys; S3 failure is non-fatal (200 + warning field); deleteJobRole has s3:DeleteObject |
| Rec 8: Document Step Functions trade-off + file issue | Done | Comment in deleteJob.ts; Issue #210 filed |
| Rec 9: shared-types DTOs | Done | `GetJobApiResponse` + `DeleteJobApiResponse` in shared-types/src/jobs.ts; dual ESM+CJS build verified |
| Rec 10: Drop ConsistentRead from getJob | Done | loadJobForUser uses eventually-consistent; rationale documented |
| Rec 11: Collapse deleteJob to single conditional DeleteItem | Done | ReturnValues: ALL_OLD eliminates prior GetItem round trip |
| Rec 12: Narrow no-console:off to smoke tests | Done | ESLint override split; 2 pre-existing __tests__/ calls fixed with targeted disable comments |
| Rec 13: no-explicit-any: document trade-off | Done | .eslintrc.cjs comment explains why 'off' is needed with --max-warnings 0 |
| Rec 14: Tighten DeleteItem assertion regex | Done | Now asserts: (a) DeleteItem appears, (b) resource matches /"JobsTable[A-Za-z0-9]*/, (c) translationRole has no DeleteItem policy |
| Rec 15: createJobLambda() helper | Done | Applied to getJob + deleteJob; existing 11 sites unchanged (minimal churn) |
| Issue 1: Soft-delete | Deferred | Issue #209 filed per user R3 decision |
| Issue 2: StopExecution | Filed | Issue #210 |
| Issue 3: Smoke as required check | Filed | Issue #211 |

## What changed (full)

| File | Change |
|---|---|
| `backend/functions/jobs/getJob.ts` | Add requestId; use loadJobForUser; use GetJobApiResponse DTO; drop ConsistentRead |
| `backend/functions/jobs/getJob.test.ts` | Add requestId to createEvent; add 2 BOLA tests; 8 tests total |
| `backend/functions/jobs/deleteJob.ts` | Add requestId; single conditional DeleteItem (CCF→404); S3 cascade; DeleteJobApiResponse DTO |
| `backend/functions/jobs/deleteJob.test.ts` | Add requestId; BOLA×2; S3 cleanup tests; ConditionalCheckFailed tests; 9 tests total |
| `backend/functions/shared/jobRepository.ts` | New: loadJobForUser() — ownership-enforcing DDB fetch |
| `backend/functions/shared/jobRepository.test.ts` | New: 5 unit tests |
| `backend/functions/jobs/getTranslationStatus.ts` | Note comment explaining why it diverges from jobRepository |
| `backend/functions/.eslintrc.cjs` | Narrow no-console:off to smoke/integration; document no-explicit-any trade-off |
| `backend/functions/chunking/__tests__/documentChunker.test.ts` | Targeted eslint-disable comment |
| `backend/functions/translation/__tests__/translateChunk.test.ts` | Targeted eslint-disable comment |
| `backend/infrastructure/lib/lfmt-infrastructure-stack.ts` | Add deleteJobRole (Role 5); createJobLambda() helper; wire deleteJob to dedicated role |
| `backend/infrastructure/lib/__tests__/infrastructure.test.ts` | Tighten DeleteItem assertion (3-part check) |
| `backend/tests/smoke/smoke.test.ts` | Add status guards before all .data.data.jobId accesses |
| `shared-types/src/jobs.ts` | Add GetJobApiResponse + DeleteJobApiResponse DTOs |

## Verification (Round 2)

- `backend/functions npm test` — **23 suites, 503 passed, 506 total** (+10 new tests)
- `backend/functions npm run lint` — **0 warnings, 0 errors**
- `backend/functions npm run format:check` — **all files pass Prettier**
- `backend/infrastructure npm run build` — **tsc exits 0**
- `backend/infrastructure npm test` — **51 passed**
- `cdk synth --context environment=dev --context skipLambdaBundling=true` — **clean**
- `shared-types npm run build` — **dual ESM+CJS build clean**

## Test plan

- [ ] CI "Run Tests" check passes on this PR
- [ ] After merge, `deploy-backend.yml` deploys CDK → smoke-tests job runs and all tests pass
- [ ] Manually verify: `GET /jobs/{jobId}` and `DELETE /jobs/{jobId}` return 200 with valid auth, 404 for wrong owner

### DIFF
--- a/backend/functions/.eslintrc.cjs
+++ b/backend/functions/.eslintrc.cjs
@@ -26,20 +26,36 @@ module.exports = {
   },
   overrides: [
     {
-      // Test files: relax rules that produce unavoidable noise in test contexts.
+      // All test files: relax no-explicit-any only.
       //
-      // `no-explicit-any`: aws-sdk-client-mock resolves with raw DynamoDB
-      // attribute maps ({S: '...', N: '...'}) that don't match the SDK return
-      // types; `as any` is the established cast pattern across all existing
-      // test files in this repo. Partial APIGatewayProxyEvent mocks also
-      // require `as any` because every optional field would need a stub.
-      //
-      // `no-console`: smoke and integration tests deliberately emit console
-      // output (✓/✗ progress lines) so CI logs show which steps passed.
-      // These are not debug leftovers — they are load-bearing diagnostics.
+      // `no-explicit-any` is turned OFF rather than 'warn' because the CI lint
+      // script runs with --max-warnings 0, so 'warn' would block CI just as
+      // 'error' would. The trade-off: developers see no editor signal for new
+      // `any` introduced in test code. Mitigation: PRs that introduce new `any`
+      // in test files should be questioned in review (this comment is the
+      // documented reminder). The cast patterns here are genuinely unavoidable:
+      //   - aws-sdk-client-mock resolves with raw DynamoDB attribute maps
+      //     ({S: '...', N: '...'}) that don't match the SDK return types.
+      //   - Partial APIGatewayProxyEvent mocks require `as any` because
+      //     supplying every optional field would bloat the test fixtures.
       files: ['**/*.test.ts', '**/__tests__/**/*.ts'],
       rules: {
         '@typescript-eslint/no-explicit-any': 'off',
+      },
+    },
+    {
+      // Smoke and integration tests only: also disable no-console.
+      //
+      // Smoke/integration tests deliberately emit console output (✓/✗ progress
+      // lines) so CI logs show which steps passed. These are load-bearing
+      // diagnostics, not debug leftovers. Unit tests do NOT get this exemption —
+      // they should use the structured Logger or Jest's built-in output instead.
+      files: [
+        '**/tests/smoke/**/*.ts',
+        '**/tests/integration/**/*.ts',
+        '**/__tests__/integration/**/*.ts',
+      ],
+      rules: {
         'no-console': 'off',
       },
     },
--- a/backend/functions/.eslintrc.cjs
+++ b/backend/functions/.eslintrc.cjs
@@ -26,6 +26,24 @@ module.exports = {
   },
   overrides: [
     {
+      // Test files: relax rules that produce unavoidable noise in test contexts.
+      //
+      // `no-explicit-any`: aws-sdk-client-mock resolves with raw DynamoDB
+      // attribute maps ({S: '...', N: '...'}) that don't match the SDK return
+      // types; `as any` is the established cast pattern across all existing
+      // test files in this repo. Partial APIGatewayProxyEvent mocks also
+      // require `as any` because every optional field would need a stub.
+      //
+      // `no-console`: smoke and integration tests deliberately emit console
+      // output (✓/✗ progress lines) so CI logs show which steps passed.
+      // These are not debug leftovers — they are load-bearing diagnostics.
+      files: ['**/*.test.ts', '**/__tests__/**/*.ts'],
+      rules: {
+        '@typescript-eslint/no-explicit-any': 'off',
+        'no-console': 'off',
+      },
+    },
+    {
       // Apply type-checking rules only to non-test files
       // Note: Strict type-checking will be fully enforced in Phase 1.1 (TypeScript Strict Mode Migration)
       files: ['**/*.ts'],
--- a/backend/functions/chunking/__tests__/documentChunker.test.ts
+++ b/backend/functions/chunking/__tests__/documentChunker.test.ts
@@ -72,6 +72,7 @@ describe('DocumentChunker', () => {
         const tokens = countTokens(chunk.primaryContent);
         expect(tokens).toBeLessThanOrEqual(3500);
         if (tokens > 3500) {
+          // eslint-disable-next-line no-console -- diagnostic: only runs when the assertion above fails
           console.error(`Chunk ${index} exceeds limit: ${tokens} tokens`);
         }
       });
--- a/backend/functions/jobs/deleteJob.test.ts
+++ b/backend/functions/jobs/deleteJob.test.ts
@@ -1,0 +1,152 @@
+/**
+ * Unit tests for Delete Job endpoint
+ * DELETE /jobs/{jobId}
+ */
+
+// Set environment variables BEFORE imports
+process.env.JOBS_TABLE = 'test-jobs-table';
+
+import { APIGatewayProxyEvent } from 'aws-lambda';
+import { mockClient } from 'aws-sdk-client-mock';
+import { DynamoDBClient, GetItemCommand, DeleteItemCommand } from '@aws-sdk/client-dynamodb';
+import { handler } from './deleteJob';
+
+const dynamoMock = mockClient(DynamoDBClient);
+
+describe('deleteJob endpoint', () => {
+  beforeEach(() => {
+    dynamoMock.reset();
+    jest.clearAllMocks();
+  });
+
+  /** Build a minimal API Gateway event for DELETE /jobs/{jobId} */
+  const createEvent = (
+    jobId: string,
+    userId = 'user-123'
+  ): Partial<APIGatewayProxyEvent> => ({
+    httpMethod: 'DELETE',
+    path: `/jobs/${jobId}`,
+    pathParameters: { jobId },
+    headers: { Authorization: 'Bearer mock-token' },
+    requestContext: {
+      authorizer: {
+        claims: { sub: userId, email: 'test@example.com' },
+      },
+    } as any,
+  });
+
+  // ---------------------------------------------------------------------------
+  // Happy-path
+  // ---------------------------------------------------------------------------
+
+  it('should delete an existing owned job and return 200', async () => {
+    // First call: GetItem (existence check)
+    dynamoMock.on(GetItemCommand).resolves({
+      Item: {
+        jobId: { S: 'job-abc' },
+        userId: { S: 'user-123' },
+        status: { S: 'PENDING_UPLOAD' },
+        createdAt: { S: '2026-01-01T00:00:00.000Z' },
+      },
+    } as any);
+
+    // Second call: DeleteItem (the actual removal)
+    dynamoMock.on(DeleteItemCommand).resolves({});
+
+    const result = await handler(createEvent('job-abc') as APIGatewayProxyEvent);
+
+    expect(result.statusCode).toBe(200);
+    const body = JSON.parse(result.body);
+    expect(body.jobId).toBe('job-abc');
+    expect(body.message).toContain('job-abc');
+  });
+
+  // ---------------------------------------------------------------------------
+  // Not found
+  // ---------------------------------------------------------------------------
+
+  it('should return 404 when job does not exist', async () => {
+    dynamoMock.on(GetItemCommand).resolves({ Item: undefined });
+
+    const result = await handler(createEvent('missing-job') as APIGatewayProxyEvent);
+
+    expect(result.statusCode).toBe(404);
+    const body = JSON.parse(result.body);
+    expect(body.message).toContain('missing-job');
+  });
+
+  // ---------------------------------------------------------------------------
+  // Auth failures
+  // ---------------------------------------------------------------------------
+
+  it('should return 401 when no authorizer claims are present', async () => {
+    const event: Partial<APIGatewayProxyEvent> = {
+      httpMethod: 'DELETE',
+      path: '/jobs/job-abc',
+      pathParameters: { jobId: 'job-abc' },
+      headers: {},
+      requestContext: {} as any,
+    };
+
+    const result = await handler(event as APIGatewayProxyEvent);
+
+    expect(result.statusCode).toBe(401);
+  });
+
+  // ---------------------------------------------------------------------------
+  // Bad request
+  // ---------------------------------------------------------------------------
+
+  it('should return 400 when jobId path parameter is missing', async () => {
+    const event: Partial<APIGatewayProxyEvent> = {
+      httpMethod: 'DELETE',
+      path: '/jobs/',
+      pathParameters: null,
+      headers: {},
+      requestContext: {
+        authorizer: { claims: { sub: 'user-123' } },
+      } as any,
+    };
+
+    const result = await handler(event as APIGatewayProxyEvent);
+
+    expect(result.statusCode).toBe(400);
+  });
+
+  // ---------------------------------------------------------------------------
+  // DynamoDB error on GetItem
+  // ---------------------------------------------------------------------------
+
+  it('should return 500 when DynamoDB GetItem fails unexpectedly', async () => {
+    dynamoMock.on(GetItemCommand).rejects(new Error('DynamoDB unavailable'));
+
+    const result = await handler(createEvent('job-abc') as APIGatewayProxyEvent);
+
+    expect(result.statusCode).toBe(500);
+    const body = JSON.parse(result.body);
+    expect(body.message).toBe('Failed to delete job');
+  });
+
+  // ---------------------------------------------------------------------------
+  // DynamoDB error on DeleteItem
+  // ---------------------------------------------------------------------------
+
+  it('should return 500 when DynamoDB DeleteItem fails unexpectedly', async () => {
+    dynamoMock.on(GetItemCommand).resolves({
+      Item: {
+        jobId: { S: 'job-abc' },
+        userId: { S: 'user-123' },
+        status: { S: 'PENDING_UPLOAD' },
+        createdAt: { S: '2026-01-01T00:00:00.000Z' },
+      },
+    } as any);
+
+    dynamoMock.on(DeleteItemCommand).rejects(new Error('DeleteItem failed'));
+
+    const result = await handler(createEvent('job-abc') as APIGatewayProxyEvent);
+
+    expect(result.statusCode).toBe(500);
+    const body = JSON.parse(result.body);
+    expect(body.message).toBe('Failed to delete job');
+  });
+});
--- a/backend/functions/jobs/deleteJob.test.ts
+++ b/backend/functions/jobs/deleteJob.test.ts
@@ -5,17 +5,25 @@
 
 // Set environment variables BEFORE imports
 process.env.JOBS_TABLE = 'test-jobs-table';
+process.env.DOCUMENT_BUCKET = 'test-document-bucket';
 
 import { APIGatewayProxyEvent } from 'aws-lambda';
 import { mockClient } from 'aws-sdk-client-mock';
-import { DynamoDBClient, GetItemCommand, DeleteItemCommand } from '@aws-sdk/client-dynamodb';
+import {
+  DynamoDBClient,
+  DeleteItemCommand,
+  ConditionalCheckFailedException,
+} from '@aws-sdk/client-dynamodb';
+import { S3Client, DeleteObjectCommand } from '@aws-sdk/client-s3';
 import { handler } from './deleteJob';
 
 const dynamoMock = mockClient(DynamoDBClient);
+const s3Mock = mockClient(S3Client);
 
 describe('deleteJob endpoint', () => {
   beforeEach(() => {
     dynamoMock.reset();
+    s3Mock.reset();
     jest.clearAllMocks();
   });
 
@@ -26,29 +34,29 @@ describe('deleteJob endpoint', () => {
     pathParameters: { jobId },
     headers: { Authorization: 'Bearer mock-token' },
     requestContext: {
+      requestId: 'test-request-id',
       authorizer: {
         claims: { sub: userId, email: 'test@example.com' },
       },
     } as any,
   });
 
+  /** DynamoDB Attributes map for a deleted job (returned via ReturnValues: ALL_OLD) */
+  const deletedJobAttributes = {
+    jobId: { S: 'job-abc' },
+    userId: { S: 'user-123' },
+    status: { S: 'PENDING_UPLOAD' },
+    s3Key: { S: 'uploads/user-123/file-001/document.txt' },
+    createdAt: { S: '2026-01-01T00:00:00.000Z' },
+  };
+
   // ---------------------------------------------------------------------------
   // Happy-path
   // ---------------------------------------------------------------------------
 
   it('should delete an existing owned job and return 200', async () => {
-    // First call: GetItem (existence check)
-    dynamoMock.on(GetItemCommand).resolves({
-      Item: {
-        jobId: { S: 'job-abc' },
-        userId: { S: 'user-123' },
-        status: { S: 'PENDING_UPLOAD' },
-        createdAt: { S: '2026-01-01T00:00:00.000Z' },
-      },
-    } as any);
-
-    // Second call: DeleteItem (the actual removal)
-    dynamoMock.on(DeleteItemCommand).resolves({});
+    dynamoMock.on(DeleteItemCommand).resolves({ Attributes: deletedJobAttributes } as any);
+    s3Mock.on(DeleteObjectCommand).resolves({});
 
     const result = await handler(createEvent('job-abc') as APIGatewayProxyEvent);
 
@@ -56,20 +64,95 @@ describe('deleteJob endpoint', () => {
     const body = JSON.parse(result.body);
     expect(body.jobId).toBe('job-abc');
     expect(body.message).toContain('job-abc');
+    expect(body.requestId).toBe('test-request-id');
+    // No warning when S3 cleanup succeeds
+    expect(body.warning).toBeUndefined();
+  });
+
+  it('should issue two S3 DeleteObject calls (uploads/ + documents/ copy)', async () => {
+    dynamoMock.on(DeleteItemCommand).resolves({ Attributes: deletedJobAttributes } as any);
+    s3Mock.on(DeleteObjectCommand).resolves({});
+
+    await handler(createEvent('job-abc') as APIGatewayProxyEvent);
+
+    const s3Calls = s3Mock.commandCalls(DeleteObjectCommand);
+    expect(s3Calls).toHaveLength(2);
+    const keys = s3Calls.map((c) => c.args[0].input.Key as string);
+    expect(keys).toContain('uploads/user-123/file-001/document.txt');
+    expect(keys).toContain('documents/user-123/file-001/document.txt');
   });
 
   // ---------------------------------------------------------------------------
-  // Not found
+  // BOLA (Broken Object Level Authorization) — OWASP API1:2023
+  // ---------------------------------------------------------------------------
+
+  it('returns 404 when job exists but is owned by a different user', async () => {
+    // The ConditionExpression (userId = :requesterUserId) fails when the stored
+    // userId does not match — DynamoDB throws ConditionalCheckFailedException.
+    // The handler must return 404 (NOT 403) to avoid leaking existence.
+    dynamoMock
+      .on(DeleteItemCommand)
+      .rejects(
+        new ConditionalCheckFailedException({ message: 'Condition check failed', $metadata: {} })
+      );
+
+    // 'attacker-999' tries to delete a job owned by 'user-123'
+    const result = await handler(createEvent('job-abc', 'attacker-999') as APIGatewayProxyEvent);
+
+    expect(result.statusCode).toBe(404);
+  });
+
+  it('does not return job data in 404 response body (BOLA privacy)', async () => {
+    // Even on the error path, the response body must not contain any job fields.
+    dynamoMock
+      .on(DeleteItemCommand)
+      .rejects(
+        new ConditionalCheckFailedException({ message: 'Condition check failed', $metadata: {} })
+      );
+
+    const result = await handler(createEvent('job-abc', 'attacker-999') as APIGatewayProxyEvent);
+
+    const body = JSON.parse(result.body);
+    expect(body).not.toHaveProperty('status');
+    expect(body).not.toHaveProperty('s3Key');
+    expect(body).not.toHaveProperty('userId');
+    expect(body).not.toHaveProperty('createdAt');
+  });
+
+  // ---------------------------------------------------------------------------
+  // Not found: job does not exist (same ConditionalCheckFailedException path)
   // ---------------------------------------------------------------------------
 
   it('should return 404 when job does not exist', async () => {
-    dynamoMock.on(GetItemCommand).resolves({ Item: undefined });
+    dynamoMock
+      .on(DeleteItemCommand)
+      .rejects(
+        new ConditionalCheckFailedException({ message: 'Condition check failed', $metadata: {} })
+      );
 
     const result = await handler(createEvent('missing-job') as APIGatewayProxyEvent);
 
     expect(result.statusCode).toBe(404);
     const body = JSON.parse(result.body);
     expect(body.message).toContain('missing-job');
+  });
+
+  // ---------------------------------------------------------------------------
+  // S3 cleanup failure (non-fatal — warning in response)
+  // ---------------------------------------------------------------------------
+
+  it('returns 200 with a warning when DDB delete succeeds but S3 cleanup fails', async () => {
+    dynamoMock.on(DeleteItemCommand).resolves({ Attributes: deletedJobAttributes } as any);
+    s3Mock.on(DeleteObjectCommand).rejects(new Error('S3 unavailable'));
+
+    const result = await handler(createEvent('job-abc') as APIGatewayProxyEvent);
+
+    // 200 — user's intent (delete the job) is honored at the DB level
+    expect(result.statusCode).toBe(200);
+    const body = JSON.parse(result.body);
+    expect(body.jobId).toBe('job-abc');
+    expect(typeof body.warning).toBe('string');
+    expect(body.warning).toContain('S3 cleanup');
   });
 
   // ---------------------------------------------------------------------------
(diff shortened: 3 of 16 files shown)


leixiaoyu:
## Performance Review (OMC) — PR #208

**Critical**:
- None. Both new endpoints use point-key DynamoDB operations against the composite PK (jobId HASH + userId RANGE) — `getJob.ts:28-37` uses `GetItemCommand`, `deleteJob.ts:38-50` uses `GetItemCommand` + `DeleteItemCommand`. No `Query`/`Scan`. Latency profile is single-digit ms p50, single-digit-RCU/WCU cost per call.

**Recommended**:
- `getJob.ts:31` and `deleteJob.ts:42` both use `ConsistentRead: true`. Strongly-consistent reads cost 2x RCU and have higher p99 latency than eventually-consistent. For a generic GET handler that polls job status, eventual consistency is almost always sufficient (clients already poll on an interval). Consider dropping `ConsistentRead` on `getJob.ts` to halve RCU cost. For `deleteJob.ts` the consistent read is more defensible (avoids 404-then-delete races) but the subsequent `DeleteItemCommand` already has `ConditionExpression: 'attribute_exists(jobId)'` (`deleteJob.ts:55`) which gives you the same race protection — the consistent GET is redundant safety. Drop it and rely on the conditional delete.
- `deleteJob.ts` performs **two round-trips** (GetItem then DeleteItem) per request. For a delete-by-id flow you can collapse into a single `DeleteItemCommand` with `ConditionExpression: 'attribute_exists(jobId) AND userId = :uid'` and `ReturnValues: 'ALL_OLD'` — one network hop instead of two, ~50% latency reduction, and the returned old item lets you preserve the `previousStatus` log line at `deleteJob.ts:101`. The current pattern is correct but wastes a round-trip.

**Suggestions**:
- API Gateway request validator: the stack creates a global `RequestValidator` at `lfmt-infrastructure-stack.ts:516` with `validateRequestParameters: true`, but the new GET/DELETE methods at lines 1762/1769 are added without `requestValidator` or `requestParameters` declarations, so path-parameter validation does not apply. Malformed jobIds (any string) reach the Lambda. Impact is small (cost is ~$0.20 per 1M Lambda invocations vs API Gateway rejecting at the edge for free) but for an auth-gated endpoint behind Cognito the practical exposure is near-zero. Optional hardening: add `requestParameters: { 'method.request.path.jobId': true }` + `requestValidator` to reject empty paths at the edge. Low priority.
- Cold start: 256MB memory + esbuild minified bundle + `@aws-sdk/*` marked external (uses Lambda-provided SDK) = expected cold-start <500ms. No provisioned concurrency needed for these infrequent endpoints — agreed.
- CloudWatch log volume: handlers emit one INFO line per request (`getJob.ts:60`, `deleteJob.ts:78`) plus one INFO on success. Default `LOG_LEVEL` is INFO (`logger.ts:41`). Volume is bounded by request rate (low) — no concern. Logger correctly suppresses DEBUG by default.
- Test runtime: 12 new unit tests use `aws-sdk-client-mock` (in-memory, zero I/O). CI duration impact <1s. Negligible.

**Positive notes**:
- IAM scope-down at `lfmt-infrastructure-stack.ts:1170-1176` uses an explicit `addToRolePolicy` for `dynamodb:DeleteItem` scoped to `jobsTable.tableArn` instead of `grantWriteData()` — avoids granting unused PutItem/UpdateItem/BatchWriteItem. Good least-privilege practice that also shrinks IAM evaluation cost (slightly).
- Both Lambdas share the same `translationRole` (line 1131, 1152) — avoids per-Lambda role creation overhead and keeps IAM resource count flat.
- Bundle config (`minify: true`, `externalModules: ['aws-sdk', '@aws-sdk/*']`) is correct — keeps deployment package small and leverages the Lambda runtime's bundled SDK for faster cold starts.
- Composite-key design (jobId + userId) means ownership enforcement is free — no extra read or filter expression needed. Optimal for this access pattern.

leixiaoyu:
## Test Coverage Review (OMC) — PR #208

**Critical**:

- **BOLA test missing in both handler suites.** Neither `getJob.test.ts` nor `deleteJob.test.ts` contains a test for the cross-ownership case: job exists in DynamoDB but `userId` on the record differs from the caller's `sub`. The implementation enforces this at the DynamoDB key level (composite HASH+RANGE key means a mis-owner `GetItem` returns no `Item`) — but there is no unit test that explicitly exercises it. A future refactor that adds a GSI or changes the key structure could silently break BOLA protection. Required test: mock `GetItemCommand` to return an `Item` whose `userId` does not match the authorizer `sub`, assert 404 (not 403, not 200). `getJob.test.ts` line 356 (happy-path block) and `deleteJob.test.ts` line 74 — both need a 7th case. Given the stated rationale for this PR ("smoke test silently broken for weeks"), a missing ownership-check unit test is the exact same class of silent-breakage risk.

- **`beforeAll` in smoke test still has unguarded `.data.data.jobId` accesses.** The bug description correctly identifies two `beforeAll` crashes, but the fix only bumps `fileSize`. The dereference chain `uploadResponse.data.data.jobId` at smoke.test.ts lines ~652 and ~704 (Job Deletion `beforeAll`) is still unguarded — if `uploadResponse.status` is anything other than 200, the property access throws `TypeError` and the entire describe-block's tests report misleading cascade failures. A status assertion before the dereference is the minimal fix: `expect(uploadResponse.status).toBe(200)` followed by `if (uploadResponse.status !== 200) throw new Error(...)`. The Translation Status Polling `beforeAll` (line ~488) has the same unguarded pattern with `uploadResponse.data.data.jobId`.

**Recommended**:

- **`no-console` override is broader than necessary.** `.eslintrc.cjs` line 21: the override applies `files: ['**/*.test.ts', '**/__tests__/**/*.ts']`, which disables `no-console` for all unit test files as well as smoke tests. The stated rationale ("smoke and integration tests deliberately emit console output") applies only to the smoke runner. Unit tests in `__tests__/` and `jobs/*.test.ts` have no load-bearing `console.log` calls — suppressing the rule there loses signal without benefit. Tighter scope: split the override into two entries — keep `no-console: off` for `**/smoke/**` only, and in the unit-test entry drop the `no-console` line entirely (or set it to `warn`).

- **`no-explicit-any` fully off vs. `warn` for unit tests.** `.eslintrc.cjs` line 23: the rule is `off`. The rationale is valid (SDK attribute maps + partial event stubs require `as any`). However, silencing the rule entirely means new test code that uses `any` out of laziness gets no signal. Setting the rule to `['warn', { ignoreRestArgs: true }]` instead of `'off'` would preserve visibility while keeping CI green — the pattern already established in the existing test files produces at most a handful of expected warnings.

- **`deleteJob` 500 response does not distinguish GetItem vs. DeleteItem failure.** `deleteJob.ts` line 309 returns `'Failed to delete job'` for both the GetItem-throws case and the DeleteItem-throws case. The two unit tests at `deleteJob.test.ts` lines 152 and 166 verify the same `body.message` string for different failure modes. An operator seeing a 500 in CloudWatch cannot distinguish a read-phase DynamoDB failure from a write-phase failure without inspecting log context. Low-effort improvement: propagate a discriminating log tag (`phase: 'get'` vs `phase: 'delete'`) in the `logger.error` call; the tests could then assert the mock was called in the expected phase.

- **DeleteItem assertion regex: `JobsTable` match does not reject wrong-table ARNs.** `infrastructure.test.ts` line 501: `expect(resourceStr).toMatch(/JobsTable/i)`. This would pass if a future policy granted DeleteItem on both the jobs table and a hypothetical `JobsTableArchive`. The guard should be strengthened to an exact logical-ID match or to assert `deleteItemStatements.length === 1` (exactly one DeleteItem grant exists). The wildcard rejection at line 498 is correct; the table-scoping check needs tightening.

- **Smoke test process discipline: no CI enforcement against long-running red tests.** The PR description correctly diagnoses that the smoke test was "low priority, not demo-critical" for weeks. There is no mechanism in `.github/workflows/` or the test config to surface this drift before it becomes a deploy blocker. Suggested lightweight control: add a `// @smoke-required` annotation convention and a CI step that fails the build if any `@smoke-required` test suite has been skipped or suppressed without a corresponding open GitHub issue reference. Alternatively, promote the smoke test job in `deploy-backend.yml` to a required check (same status as `Run Tests`) so a red smoke cannot be merged away — it must be fixed or explicitly suppressed via issue number.

**Suggestions**:

- **Shared `createEvent` helper between `getJob.test.ts` and `deleteJob.test.ts`.** Both files define an identically-structured local `createEvent` helper (lines 58 and 340 of the diff). This is low-risk duplication now but will drift if the `APIGatewayProxyEvent` mock shape changes. A shared `tests/fixtures/api-event.ts` factory would eliminate the duplication and make the shape change a one-file edit.

- **Happy-path response in `deleteJob` is 200 with a body, but smoke test at line ~720 accepts `[200, 204]`.** `deleteJob.ts` line 298 always returns 200. The smoke test's `expect([200, 204]).toContain(response.status)` is slightly over-permissive — 204 would fail the subsequent body assertions in the same test if the handler ever switches to 204 No Content. The unit test correctly pins to 200; the smoke test should match.

- **`tokenizer.ts` `eslint-disable-next-line` is the one borderline case.** `logger.ts` lines 617 and 621: `console.error`/`console.log` are the terminal output path for the structured logger — the suppress is clearly correct. `tokenizer.ts` line 637: the rationale says "no Logger dependency; fallback warning must reach CloudWatch" but `console.warn` in a Lambda that already imports `Logger` is a design smell. The correct fix for a future cleanup pass is to add a Logger instance to `tokenizer.ts` (it's a tiny addition) and remove the suppress entirely. The suppress is acceptable for this PR's scope; flagging it here for tracking.

**Positive notes**:

- Mocks are at the right layer: `mockClient(DynamoDBClient)` with `aws-sdk-client-mock` intercepts at the SDK v3 client level, not by monkey-patching the module. Both `getJob.test.ts:331` and `deleteJob.test.ts:49` use this correctly.
- Raw DynamoDB attribute-map format (`{S: '...', N: '...'}`) in mock responses is realistic — it matches what `GetItemCommand` actually returns before `unmarshall`, so the tests exercise the full deserialization path.
- `EXPECTED_APPLICATION_LAMBDA_COUNT 11→13` is correctly pinned and the comment at `infrastructure.test.ts:1433` names all 13 Lambdas explicitly — future additions cannot silently pass.
- The `ConditionExpression: 'attribute_exists(jobId)'` in `deleteJob.ts:252` correctly guards against the TOCTOU race between `GetItem` and `DeleteItem`. The DeleteItem-error test at `deleteJob.test.ts:166` covers the failure path of this path.
- ESLint override rationale comments are well-written and reference the concrete SDK patterns that force the relaxation — this is the right level of documentation for a team rule change.

leixiaoyu:
## Architecture Review (OMC) — PR #208

**Critical**:
- None.

**Recommended**:
- **Cascade/orphaning is unhandled.** `DELETE /jobs/{jobId}` removes only the DynamoDB row. Per `lfmt-infrastructure-stack.ts:883-884` the same `documentBucket` stores both source documents and chunks, and active Step Functions executions are not stopped (`backend/functions/jobs/deleteJob.ts:200-203` documents this trade-off). The handler's own header docstring acknowledges orphaning. Architecturally the contract is incoherent: a 200 OK leaves S3 objects (source + N chunk files) and potentially a running SFN execution that will continue writing to the now-orphaned `jobId`. Three coherent options, in order of architectural cost: (1) refuse delete while `status IN_PROGRESS` (return 409 Conflict) and require user to cancel first; (2) implement soft-delete (`deletedAt` marker + lifecycle/scheduled purge of S3 + DDB), which is the standard pattern for user-deletable content potentially under legal-attestation retention; (3) full cascade (S3 prefix delete + `StopExecution`). For a translation service that asks users to attest copyright ownership at upload (`smoke.test.ts:642-648`), soft-delete is almost certainly the right long-term answer — flag this as an explicit product decision before this endpoint gains a real frontend caller.
- **No `shared-types` DTOs for the new endpoints.** `shared-types/src/jobs.ts` already houses `CreateJobResponse`, `JobProgressResponse`, `CancelJobResponse`, etc. (lines 99/132/268), but `getJob.ts:507-518` defines `GetJobResponse` as a private interface inside the Lambda, and `deleteJob.ts` returns an ad-hoc `{ message, jobId }` shape with no contract. Frontend `translationService.ts` cannot type-safely call these endpoints. Add `GetJobResponse` + `DeleteJobResponse` (+ `zod` schemas to match the file's existing pattern at `jobs.ts:277`) so the contract lives at the boundary.
- **YAGNI smell on `GET /jobs/{jobId}` shape.** The PR comment on `getJob.ts:481-484` admits the response is a flat object specifically so a smoke-test assertion resolves — this is the test wagging the dog. The richer endpoint `GET /jobs/{jobId}/translation-status` already returns a superset (`getTranslationStatus.ts:99-123`). Either (a) delete `getJob.ts` and have the smoke test continue to use `/translation-status` (it already does after Bug 1's fix at `smoke.test.ts:520-531`), or (b) justify `GET /jobs/{jobId}` as an intentional minimal/cheap projection and document why — but currently nothing in the production code path consumes it (verified no `frontend/src/**` caller).

**Suggestions**:
- **CDK monolith stress signal — promote #64.** `lfmt-infrastructure-stack.ts` is now 2,096 LOC (`wc -l`) with 13 inline `NodejsFunction` declarations and all API Gateway wiring (`createApiEndpoints` at line 1602 onward). The duplicated `bundling`/`runtime`/`architecture`/`role` block in `getJobFunction` and `deleteJobFunction` (`:1121-1161`) is a textbook case for a `JobLambda` L3 construct or per-domain nested stack. Issue #64 is correctly scoped — this PR is fresh evidence to bump its priority. Lightweight intermediate step: a private `createJobLambda(id, entry, description, extraPolicies?)` helper in this stack would remove ~30 LOC of duplication today without waiting for the nested-stack refactor.
- **Two new `*.test.ts` files have ~80% scaffolding duplication** (`createEvent` builder, `dynamoMock` setup, identical 401/400/500 cases). Extract `backend/functions/jobs/__tests__/test-helpers.ts` exporting `createAuthedEvent(method, jobId, userId?)`, `setupDynamoMock()`, and shared "auth/path-param/DDB-error" assertion factories. Will pay off on the next jobs endpoint and reduce drift between handlers.
- **Smoke-test drift discipline (architectural):** the report that smoke was silently broken for weeks (test polled an endpoint that never existed; the team treated red as ambient noise) is a process-architecture gap, not a code gap. Worth filing a follow-up: either (a) make the smoke job a required check on `main` (forces immediate triage), or (b) require any failing-but-tolerated check to carry an explicit TODO-with-issue comment in the workflow file. Quietly-red required checks corrode all later signal.
- **ESLint test override is appropriately bounded** (`.eslintrc.cjs:28-44`: only `**/*.test.ts` + `__tests__/**`, only `no-explicit-any` and `no-console`, with multi-paragraph rationale). No slippery-slope concern. Suggest one tightening: include `**/tests/**/*.ts` glob too (smoke tests under `backend/tests/smoke/` may not match `__tests__/`) so the comment's "smoke and integration tests" claim actually applies.

**Positive notes**:
- New handlers correctly mirror the established pattern from `getTranslationStatus.ts`: Cognito-claims auth-check, `createSuccessResponse`/`createErrorResponse`, `Logger` (not raw `console`), `getRequiredEnv`, `marshall`/`unmarshall`, `ConsistentRead: true`, `DynamoDBJob` from `shared-types`. Pattern coherence is high.
- IAM choice is architecturally correct: `addToRolePolicy` with explicit `dynamodb:DeleteItem` scoped to `this.jobsTable.tableArn` (`lfmt-infrastructure-stack.ts:1153-1159`) preserves least-privilege and the matching infra test (`infrastructure.test.ts:451-481`) was rewritten from "no DeleteItem anywhere" to "DeleteItem must be scoped to JobsTable" — that's the right way to evolve a security invariant rather than weaken it.
- `DeleteItemCommand` uses `ConditionExpression: 'attribute_exists(jobId)'` (`deleteJob.ts:252`) to guard against concurrent-delete races. Good defensive design.
- `EXPECTED_APPLICATION_LAMBDA_COUNT` 11→13 update + comment (`infrastructure.test.ts:1436-1441`) keeps the count guard in sync. Boy-scout discipline preserved.
- The 3 `eslint-disable-next-line no-console` in `logger.ts:67/70` and `tokenizer.ts:27` carry CloudWatch rationale comments. Targeted, justified, not blanket.

leixiaoyu:
## Security Review (OMC) — PR #208

**Critical**:
- None. The destructive endpoint is correctly gated by Cognito + composite-key ownership at the data layer.

**Recommended**:
- **Defense-in-depth: enforce ownership in the IAM policy too.** `addToRolePolicy` at `lfmt-infrastructure-stack.ts:789-795` grants `dynamodb:DeleteItem` on the entire `jobsTable.tableArn` with no `Condition`. The composite-key check is the only barrier — a future code bug that constructs `Key: { jobId, userId: someOtherUserId }` would silently delete another user's row. Add an IAM `Condition` such as `dynamodb:LeadingKeys = ${aws:PrincipalTag/userId}` (or similar via STS session tags) — or at minimum a CDK comment documenting the layered enforcement contract so future contributors don't break the invariant.
- **Inline policy leaks DeleteItem to every Lambda on `translationRole`.** `deleteJob`/`getJob` reuse the shared `translationRole` (`lfmt-infrastructure-stack.ts:749, 772`), and `addToRolePolicy` mutates that shared role. So `translateChunk`, `startTranslation`, `getTranslationStatus`, *and* every legacy Lambda (auth, upload, chunking) that still aliases `lambdaRole = translationRole` (`lfmt-infrastructure-stack.ts:842`) all gain `dynamodb:DeleteItem` on the jobs table. The PR description claims least-privilege ("scoped inline policy rather than broadening the shared translationRole" — `lfmt-infrastructure-stack.ts:763-764`), but functionally the blast radius is every Lambda on that role. Fix: give `deleteJobFunction` its own dedicated execution role (mirror the `authRole`/`uploadRole`/`chunkingRole` pattern referenced at `lfmt-infrastructure-stack.ts:841`) and attach the DeleteItem policy *only* to that role.
- **Orphaned data on DELETE.** `deleteJob.ts:184-256` removes the DynamoDB row but does not delete the corresponding S3 objects (`documents/${userId}/${fileId}/...` in `documentBucket`, plus any results in `resultsBucket`) — the file header comment only acknowledges the in-flight Step Function trade-off (`deleteJob.ts:198-203`), not the storage one. From a privacy perspective this is the more impactful gap: a user who clicks "Delete my job" reasonably expects their uploaded copyrighted content to be gone. Either (a) cascade S3 deletes (DeleteObjectsByPrefix on `${userId}/${fileId}/`) or (b) document the retention behavior in the API response and a privacy notice. GDPR/right-to-erasure exposure if this ever ships beyond POC.
- **No StopExecution on in-flight Step Functions.** Acknowledged in the file header (`deleteJob.ts:198-203`) but worth tracking — a deleted job whose Step Function is still running will continue to consume Gemini quota, write progress to a now-deleted DynamoDB row (silent UpdateItem failures), and bill compute. Add a follow-up issue + a `TODO` linking to it.

**Suggestions**:
- **404 vs 403 for not-owned jobs — already correct.** The composite-key `GetItem` returns `null` whether the jobId doesn't exist OR exists under another userId, and `getJob.ts:565-573` / `deleteJob.ts:282-290` both return the same 404 with message `"Job not found: ${jobId}"`. No existence-leak. (Positive — call out below.)
- **Per-user/per-route throttling.** API Gateway stage throttle is 100 RPS dev / 1000 RPS prod (`lfmt-infrastructure-stack.ts:498-499`) — that's account-wide, not per-caller. A single authenticated user could loop `DELETE /jobs/{jobId}` against their own jobs at full burst. Consider an APIGW UsagePlan keyed off the Cognito `sub`, or a per-method MethodOptions throttle on the DELETE specifically (lower than read endpoints).
- **Audit log line on DELETE could include source IP + requestId.** `deleteJob.ts:295` logs `{ jobId, userId, previousStatus }` — good. Add `event.requestContext.identity?.sourceIp` and `event.requestContext.requestId` so a forensic query can correlate to the API Gateway access log row.
- **Consistent reads for the existence check — confirm necessity.** `loadJob` uses `ConsistentRead: true` (`deleteJob.ts:231, getJob.ts:528`). For a destructive endpoint, fine. For `getJob` (poll-friendly read), this doubles RCU cost — eventually-consistent would be cheaper and the only race is "user just-uploaded sees stale 404", which is benign.
- **Test-file ESLint override.** Disabling `no-explicit-any` + `no-console` for `**/*.test.ts` (`.eslintrc.cjs:21-25`) is pragmatic but does sweep up the legitimate guardrail. Concretely: a future test that mocks an auth token or sets `requestContext.authorizer.claims` to a malformed shape via `as any` would no longer be flagged. Prefer per-line `eslint-disable` with the same justification, or scope the override only to `aws-sdk-client-mock` resolves (e.g. via a typed test helper). Not blocking.
- **logger.ts / tokenizer.ts `eslint-disable-next-line no-console` rationales.** Both are accurate (`logger.ts:617, 620` is the structured-logging boundary; `tokenizer.ts:636` has no Logger dependency by design). Acceptable.

**Positive notes**:
- **Ownership enforcement via composite key is the right pattern.** JobsTable is `partitionKey: jobId, sortKey: userId` (`lfmt-infrastructure-stack.ts:181-182`); `getJob.ts:524-538` and `deleteJob.ts:227-241` build `Key: marshall({ jobId, userId })` from `claims.sub`. Mismatched ownership → no row returned → uniform 404. **This is the textbook OWASP API1:2023 (BOLA) defense and it's implemented correctly.** No app-layer `if (job.userId !== requesterUserId)` check needed because the DB query itself encodes the check.
- **Uniform 404 message prevents existence enumeration.** Both "doesn't exist" and "owned by someone else" return identical `"Job not found: ${jobId}"` — no info-leak. (`getJob.ts:567`, `deleteJob.ts:285`)
- **`ConditionExpression: 'attribute_exists(jobId)'` on DeleteItem** (`deleteJob.ts:252`) makes the delete idempotent-ish: a race where two clients delete the same row at the same time → second one fails with `ConditionalCheckFailedException`, which returns 500 (could be polished to 404, but at least it doesn't lie about success).
- **CDK regression test (`infrastructure.test.ts:451-700`)** rewrites the previous "no DeleteItem anywhere" guard into a structural assertion that DeleteItem is scoped to JobsTable only and never targets `"*"` or other tables. This is exactly the right test evolution for a security-sensitive grant — it locks in the scoping invariant against future drift.
- **Authentication checks are correctly fail-closed.** Both handlers return 401 *before* touching DynamoDB if `claims.sub` is missing (`getJob.ts:551-554`, `deleteJob.ts:269-272`). Defense-in-depth complement to the Cognito authorizer.
- **API method wiring uses `AuthorizationType.COGNITO` + `authorizer`** on both new methods (`lfmt-infrastructure-stack.ts:817-818, 824-825`) — no anonymous access path exists.
- **No PII in error responses.** 500 path returns generic `"Failed to delete job"` / `"Failed to get job"` (`getJob.ts:599`, `deleteJob.ts:309`); the underlying error is logged server-side only.

leixiaoyu:
## Code Quality Review (OMC) — PR #208

**Critical** (must fix):

- **Missing `requestId` in all new Lambda responses** — `getJob.ts` and `deleteJob.ts` pass `undefined` as `requestId` to every `createErrorResponse` / `createSuccessResponse` call. Every other handler in the codebase extracts `event.requestContext.requestId` at the top of the handler and threads it through (see `login.ts:28`, `uploadRequest.ts:37`). Without it, 500-class errors from these two endpoints cannot be correlated to CloudWatch logs, which directly harms incident response. Fix: add `const requestId = event.requestContext.requestId;` at the top of each handler and pass it to all response helpers.

**Recommended**:

- **`ConditionalCheckFailedException` swallowed as 500** — `deleteJob.ts:248-256` uses `ConditionExpression: 'attribute_exists(jobId)'` to guard against a concurrent-delete race, but the outer catch at line 303 catches all errors and returns a generic 500. A `ConditionalCheckFailedException` thrown by DynamoDB in that race scenario should map to 404 (the item is already gone) or 409. Currently it surfaces as "Failed to delete job" with a 500, misleading the caller. Add a named-error check: `if (error instanceof Error && error.name === 'ConditionalCheckFailedException') return createErrorResponse(404, ...)`.

- **`loadJob` duplicated verbatim between `getJob.ts` and `deleteJob.ts`** — Both files contain an identical `loadJob(jobId, userId)` function body (composite-key GetItem with ConsistentRead). As the number of job-scoped endpoints grows, this will drift. Extract to `backend/functions/jobs/jobRepository.ts` (or similar) exporting `loadJob`, `deleteJobRecord`, etc. `getTranslationStatus.ts` already has its own copy — this is the third instance of the same function.

- **`logger.warn` never called in new handlers** — `uploadRequest.ts:48-55` and `login.ts:40` emit `logger.warn` for auth/validation failures (important for security dashboards). Both new handlers go directly to `createErrorResponse` for 401/400 without a warn log. At minimum, the 401 path should warn so Cognito misuse is visible in CloudWatch.

**Suggestions**:

- **`loadJob` placed before `handler` export, inconsistent with project style** — Existing handlers (`getTranslationStatus.ts:63,167`) define the exported `handler` first, with helper functions after (which works because TypeScript/V8 hoists `function` declarations). The new files define `loadJob` as `async function` before the `handler` arrow. Neither is wrong, but it is inconsistent. Low priority.

- **ESLint `no-explicit-any: off` for all test files** — The justification in `.eslintrc.cjs:10-16` is accurate (aws-sdk-client-mock DynamoDB attribute maps, partial `APIGatewayProxyEvent` mocks). The tradeoff is real: suppressing the rule blanket-style makes future `any` casts in test code invisible. An alternative would be scoped `// eslint-disable-next-line` at each cast site, which is more surgical. For a POC with a lint gate that was shipping 260 warnings, the blanket approach is a pragmatic call — just document it in the PR so future reviewers know why it's there (it is documented in the comment block, which is good).

- **`getJob.ts:589-592` comment explains a now-redundant design decision** — The comment "Return the response as a flat object (no `data` wrapper) so that the smoke-test assertion `response.data.jobId` resolves without an extra level of nesting" is accurate for `getJob`, but the smoke test's "poll job status" test (`smoke.test.ts:522`) was *also* changed in this PR to use `/translation-status` — so the smoke test no longer hits `getJob` for that assertion. The flat-object design is still the right call for `GET /jobs/{jobId}`, but the comment's stated motivation is stale. Update to: "Returns a flat object (no `data` envelope) consistent with REST convention for a single-resource GET."

- **Smoke test: no assertion on `response.status` before accessing `.data.data.jobId` in `beforeAll`** — The original brittleness that caused `Bug 2` (accessing `.data.data.jobId` when status was 400) is fixed by bumping `fileSize` to 1024, but the underlying access pattern is still unguarded. The comment at `smoke.test.ts:642-647` explains *why* 1024, which is good. An additional `expect(uploadResponse.status).toBe(200)` before the `jobId =` assignment would make future validator changes immediately obvious rather than producing a confusing TypeError.

**Positive notes**:

- Ownership enforcement is correct and follows the established pattern: the composite DynamoDB key (`jobId` HASH + `userId` RANGE) means a GetItem miss for the wrong user returns `Item: undefined`, which maps cleanly to 404 without leaking job existence. This is consistent with `getTranslationStatus.ts:85-91`.
- `deleteJob.ts:248-256` using an explicit `ConditionExpression` on `DeleteItem` rather than a blind delete is a good defensive pattern even if the exception isn't handled distinctly yet.
- IAM scoping: using `addToRolePolicy(dynamodb:DeleteItem)` instead of `grantWriteData()` is exactly right and the infrastructure test upgrade (`infrastructure.test.ts:651-700`) properly validates scoping rather than just absence.
- ESLint disable comments in `logger.ts:617,621` and `tokenizer.ts:636` are accurate and durable — the CloudWatch rationale is load-bearing and the comments will not become stale.
- Commit messages are exemplary: root-cause classification (B+C), hypothesis tracing, and verification checklist make the history self-documenting.

xlei-raymond (approved):
Detailed Review Findings

  1. Structural Alignment & API Completeness (Bug #1 & #3)
  The PR correctly diagnoses the 403 errors and missing route failures in CI.
   * The Problem: The smoke tests were attempting to use GET /jobs/{jobId} and DELETE /jobs/{jobId}, but these standard REST endpoints had never been implemented or wired into API Gateway.
   * The Fix: Full implementation of getJob.ts and deleteJob.ts, correctly wired into CDK.
   * Best Practice: The extraction of the jobRepository.ts helper is a major quality win. It centralizes the ownership-enforcement logic, ensuring that BOLA (Broken Object Level Authorization) is handled consistently
     across all job-related Lambdas.

  2. Security & Least Privilege (OMC Round 2 Critical 2 & 3)
  The team has demonstrated a "zero-short-cuts" security posture in this PR.
   * Role Isolation: Creating a dedicated DeleteJobLambdaRole (Role 5) that carries the only dynamodb:DeleteItem permission in the entire stack is a perfect implementation of least-privilege.
   * BOLA Verification: I highly commend the addition of the cross-ownership tests. Asserting that one user cannot GET or DELETE another user's job is a mandatory high-standard guard for multi-tenant systems.
   * Conditional Deletes: The transition to a single-round-trip conditional DeleteItem (Rec 5/11) is an excellent optimization that also eliminates a major race condition between read and delete.

  3. Forensic Integrity (Critical 1)
   * Request Correlation: Verified that both new handlers correctly extract requestId from the API Gateway context and pass it through to the structured logging and success-response layers. This preserves our end-to-end
     traceability standard.

  4. Operational Resilience (Bug #2 & Critical 4)
   * Validation Floor: Correctly identified and fixed the MIN_FILE_SIZE violation in the smoke test data.
   * Defensive Assertions: The addition of expect(uploadResponse.status).toBe(200) guards before accessing .data.data.jobId is exactly the kind of "fail-fast" test engineering we need to avoid confusing TypeError crashes
     in CI.

  5. Verification Results
  I have independently verified the implementation within the worktree:
   * Infrastructure Tests: All 51 tests pass. The new 3-part DeleteItem assertion correctly locks in the least-privilege requirement.
   * Backend Tests: All 503 backend tests pass, including the 10 new BOLA and repository scenarios.
   * Documentation: The addition of GetJobApiResponse and DeleteJobApiResponse DTOs to shared-types maintains our Single Source of Truth standard.

  Verdict: APPROVE (GO)
  This PR is 100% verified and meets our highest standards. It is a transformative upgrade to our backend reliability and security posture, definitively clearing the path for a green deployment pipeline.

  ---

  PR Review for GitHub

  Status: ✅ Approved

  This PR represents an exceptional resolution to multiple layers of post-deploy verification failures. By correctly implementing the missing REST primitives (GET and DELETE), establishing a centralized,
  ownership-enforcing repository layer, and hardening our IAM security posture, the team has restored full visibility and trust to our automated deployment pipeline.

   1. API Surface Reconciliation: The implementation of GET /jobs/{jobId} and DELETE /jobs/{jobId} resolves a long-standing contract mismatch between our smoke tests and our deployed infrastructure. The use of flat DTOs
      (GetJobApiResponse) ensures optimal frontend developer experience.
   2. Zero-Tolerance Security (BOLA Closure): I highly commend the introduction of the jobRepository layer. Centralizing the partition+range key lookup ensures that Object-Level Authorization is enforced by the database
      schema itself, effectively eliminating the risk of cross-user data leakage.
   3. Gold-Standard Least Privilege: The creation of the isolated DeleteJobLambdaRole is a textbook display of defensive infrastructure engineering. By ensuring that dynamodb:DeleteItem is restricted to exactly one
      function, we have significantly hardened the project's data-integrity surface.
   4. Optimized Deletion Protocol: The transition to a single-round-trip conditional delete with ReturnValues: ALL_OLD is a sophisticated improvement. It provides atomicity and performance gains while correctly surfacing
      404s for non-existent records.

  Final Verification Results
   * Logic Integrity: 100% (BOLA guards and request-threading verified).
   * Least Privilege: Confirmed (Regression test guards Role 5 isolation).
   * Standards Compliance: 100% (Clean lint, types, and documentation).

  ---
  Status: 100% Logic Verified; Security Surface Hardened; CI Pipeline Unblocked.

PR #203, opened 2026-05-05
chore: Node 22 LTS upgrade + shared-types ESM + Graviton + audit hardening

leixiaoyu:
## Why now

Three Node-runtime surfaces had been drifting past their support windows. This PR addresses all three plus three follow-up issues that surfaced during review (ESM migration, ARM64 Graviton, audit triage):

| Surface | Before | After | Status |
|---|---|---|---|
| AWS Lambda runtime (CDK `NODEJS_18_X`, 12 sites) | Node 18 | **Node 22** | Node 18 EOL'd April 2025 — over a year ago |
| CI runners (`NODE_VERSION` in 3 workflows) | Node 20 | **Node 22** | Node 20 just hit EOL on 2026-04-30 |
| Root `package.json` engines | `>=18.0.0` / `>=8.0.0` | **`>=22.11.0` / `>=10.9.0`** | Was permitting two EOL versions; floor now at Node 22 LTS baseline |
| `shared-types` package format | CJS only | **dual ESM + CJS** | Unblocks long-failing `vite build` CI step |
| Lambda CPU architecture | x86_64 default | **ARM64 / Graviton** | ~20% Lambda cost reduction |
| npm audit (4 packages) | 1 critical + 24 high | **0 critical, 0 high** | All four packages pass `--audit-level=high` |
| Local dev | Node 22 | (unchanged) | Already on the active LTS |

Closes #204
Closes #205
Closes #206

---

### 1. CI workflows (`.github/workflows/`)
- `ci.yml`, `deploy-backend.yml`, `deploy-frontend.yml` — bumped `NODE_VERSION: '20'` → `NODE_VERSION: '22'`. All `actions/setup-node@v4` blocks already reference `${{ env.NODE_VERSION }}`, so this is a one-line change per file.

### 2. CDK Lambda runtime (`backend/infrastructure/lib/lfmt-infrastructure-stack.ts`)
- All 12 `lambda.Runtime.NODEJS_18_X` occurrences replaced with `lambda.Runtime.NODEJS_22_X`.
- **Boy-scouts cleanup (DRY)**: extracted the runtime version to a single module-level `LAMBDA_RUNTIME` constant with a comment cross-referencing the CI `NODE_VERSION` env var and root `engines.node` constraint.

### 3. Root `package.json` engines + `.npmrc`
- `engines.node`: `>=18.0.0` → `>=22.11.0` (Node 22 LTS shipped at 22.11)
[...]

PR #212, opened 2026-05-07
fix(demo): download endpoint + smoke fixture + TranslationTone SSoT

leixiaoyu:
## Summary

Three demo-readiness fixes shipped as one PR, each in its own logical commit so they can be reviewed or bisected independently.

### Issue 1 — HIGH (demo blocker): \`GET /jobs/{jobId}/download\` endpoint

The frontend (\`translationService.ts:347\`) has been calling this endpoint since it was wired up, but no Lambda or API Gateway route existed — every call returned 403.

**What was added:**
- \`backend/functions/translation/downloadTranslation.ts\` — Lambda handler that paginates \`ListObjectsV2\`, sorts chunks numerically (not lexicographically), fetches all chunks in parallel with \`Promise.all\`, and returns a raw \`text/plain\` blob.
- Auth via Cognito authorizer; ownership via DynamoDB composite key (BOLA prevention, 404 for cross-user access).
- 409 Conflict for non-COMPLETED jobs.
- \`Content-Disposition: attachment; filename="translated_<original>"\`
- Dedicated \`DownloadTranslationLambdaRole\` scoped to \`dynamodb:GetItem\` on JobsTable, \`s3:GetObject\` on \`translated/*\` prefix only, \`s3:ListBucket\`.
- 21 unit tests + 5 infrastructure drift-guard tests; \`EXPECTED_APPLICATION_LAMBDA_COUNT\` bumped 13 → 14.

### Issue 2 — MEDIUM: Smoke fixture below MIN_FILE_SIZE

\`smoke-test-minimal.txt\` was 976 bytes; the backend rejects uploads below 1000 bytes. New size: **1354 bytes**.

### Issue 3 — LOW: \`TranslationTone\` single source of truth

Added \`TranslationTone\` type + \`TRANSLATION_TONE_VALUES\` to \`shared-types/src/jobs.ts\` and wired into \`startTranslation.ts\`, \`translateChunk.ts\`, and \`TranslationConfig.tsx\`.

---

### Critical items

| # | Item | Status |
|---|------|--------|
| 1 | BOLA test non-vacuous | Done — per-key DDB mock; test FAILs if ownership check deleted |
| 2 | Tone contract test live + SSoT fully wired | Done — drives actual startTranslation handler; stale comment was the confusion, inline was already gone |
| 3 | E2E intercept URL fix | Done — /v1/jobs/{id}/download; removed swallowing try/catch |
| 4 | URL convention: /jobs/{jobId}/download | Done — all consumers updated (translationService, MSW, handlers test, CDK, infra test, E2E) |
| 5 | TranslationConfig.tsx TONE_OPTIONS derives from TRANSLATION_TONE_VALUES | Done — TONE_METADATA Record<TranslationTone,...> + map(); ToneCode = TranslationTone; 3 contract tests |

### Recommended items

| # | Item | Status |
|---|------|--------|
| 6 | Remove private errorResponse(), use shared createErrorResponse() | Done |
| 7 | Stale comment in toneContract.test.ts rewritten | Done |
| 8 | TRANSLATION_TONE_VALUES uses as const satisfies pattern | Done |
| 9 | Extract corsPreflightOptions() helper — 3 duplicated CORS blocks replaced | Done |
| 10 | Content-Disposition filename sanitization via SAFE_FILENAME_PATTERN | Done — falls back to translated_document.txt |
| 11 | Chunk count integrity check vs job.totalChunks | Done — returns 500 with expected/got counts |
| 12 | Cap corrupt-keys log to first 10 | Done — MAX_CORRUPT_KEYS_LOGGED |
| 13 | UUID validation on jobId path param | Done — test covers path-traversal input + valid UUID |
| 14 | 6 MB body-size guard (MAX_RESPONSE_BYTES) | Done — 413 with note pointing to future presigned-URL |
| 15 | S3 client keep-alive + maxSockets: 50 | Done — NodeHttpHandler + HttpAgent |
| 16 | DownloadTranslationApiResponse marked @internal | Done |
| 17 | 409 test in translationService.test.ts | Done |
| 18 | Tighten infra test IAM assertion | Existing assertions are already tight (unchanged) |
| 19 | Multi-chunk ordering edge case chunk-0..chunk-12 | Done — 13-key test forcing two digit-length groups |
| 20 | CDK logical-ID convention comment in infra test | Done |

### Opportunistic suggestions applied
- handlers.test.ts full-happy-path URL updated to /jobs/:id/download (caught during round-2 sweep)
- @internal on DownloadTranslationApiResponse

### Deferred (out of scope)
- Presigned-URL refactor for >6 MB docs — 413 + message now points to this; follow-up issue to file
- Nested stacks + reusable scoped-role helper — genuine but defer per OMC guidance

## Test plan

- [ ] All backend function tests pass — 537 pass (3 pre-existing skips)
- [ ] All infrastructure tests pass — 57 pass
- [ ] CDK synth succeeds
- [ ] Frontend type-check, lint, format, 686 unit tests, production build all pass
- [ ] After deploy: upload smoke fixture — confirm no 400 on size
- [ ] After deploy: complete a translation, click Download — file downloads as \`translated_*.txt\`
- [ ] After deploy: attempt download on IN_PROGRESS job — confirm 409, not 500
- [ ] After deploy: attempt download as different user — confirm 404 (BOLA guard)
- [ ] After deploy: attempt download with non-UUID jobId — confirm 400

### DIFF
--- a/backend/functions/jobs/startTranslation.ts
+++ b/backend/functions/jobs/startTranslation.ts
@@ -254,7 +254,7 @@ function validateRequest(body: StartTranslationRequest): {
   // TRANSLATION_TONE_VALUES is the canonical allowlist from shared-types — the
   // single source of truth that also types the frontend TONE_OPTIONS selector.
   // Using it here ensures the backend and UI never drift independently.
-  if (body.tone && !(TRANSLATION_TONE_VALUES as ReadonlyArray<string>).includes(body.tone)) {
+  if (body.tone && !(TRANSLATION_TONE_VALUES as readonly string[]).includes(body.tone)) {
     return {
       valid: false,
       error: `Invalid tone: ${body.tone}. Must be one of: ${TRANSLATION_TONE_VALUES.join(', ')}`,
--- a/backend/functions/jobs/startTranslation.ts
+++ b/backend/functions/jobs/startTranslation.ts
@@ -13,7 +13,7 @@ import {
 } from '@aws-sdk/client-dynamodb';
 import { SFNClient, StartExecutionCommand, StartExecutionCommandOutput } from '@aws-sdk/client-sfn';
 import { marshall, unmarshall } from '@aws-sdk/util-dynamodb';
-import { DynamoDBJob } from '@lfmt/shared-types';
+import { DynamoDBJob, TRANSLATION_TONE_VALUES, TranslationTone } from '@lfmt/shared-types';
 import Logger from '../shared/logger';
 import { getRequiredEnv } from '../shared/env';
 import { createSuccessResponse, createErrorResponse } from '../shared/api-response';
@@ -54,7 +54,9 @@ const getStateMachineArn = async (): Promise<string> => {
  */
 interface StartTranslationRequest {
   targetLanguage: string;
-  tone?: 'formal' | 'informal' | 'neutral';
+  // TranslationTone is the canonical type from shared-types — single source
+  // of truth shared with the frontend TONE_OPTIONS selector.
+  tone?: TranslationTone;
   contextChunks?: number;
 }
 
@@ -249,10 +251,13 @@ function validateRequest(body: StartTranslationRequest): {
     };
   }
 
-  if (body.tone && !['formal', 'informal', 'neutral'].includes(body.tone)) {
+  // TRANSLATION_TONE_VALUES is the canonical allowlist from shared-types — the
+  // single source of truth that also types the frontend TONE_OPTIONS selector.
+  // Using it here ensures the backend and UI never drift independently.
+  if (body.tone && !(TRANSLATION_TONE_VALUES as ReadonlyArray<string>).includes(body.tone)) {
     return {
       valid: false,
-      error: `Invalid tone: ${body.tone}. Must be one of: formal, informal, neutral`,
+      error: `Invalid tone: ${body.tone}. Must be one of: ${TRANSLATION_TONE_VALUES.join(', ')}`,
     };
   }
 
--- a/backend/functions/translation/downloadTranslation.test.ts
+++ b/backend/functions/translation/downloadTranslation.test.ts
@@ -1,6 +1,6 @@
 /**
  * Unit tests for the Download Translation endpoint
- * GET /translation/{jobId}/download
+ * GET /jobs/{jobId}/download
  */
 
 // Set required environment variables BEFORE any imports so that getRequiredEnv()
@@ -25,14 +25,20 @@ function makeS3Stream(content: string) {
   return sdkStreamMixin(Readable.from([content]));
 }
 
+// Standard test UUIDs — all test-event defaults use these so the UUID
+// validation guard in the handler doesn't reject before we get to the logic
+// under test.
+const TEST_JOB_ID = '11111111-1111-1111-1111-111111111111';
+const TEST_USER_ID = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
+
 /** Build a minimal COMPLETED DynamoDB job item. */
 function makeCompletedJobItem(overrides: Record<string, unknown> = {}) {
   // removeUndefinedValues: true prevents marshall from throwing when an
   // override explicitly sets a field to undefined (e.g., { filename: undefined }).
   return marshall(
     {
-      jobId: 'job-123',
-      userId: 'user-abc',
+      jobId: TEST_JOB_ID,
+      userId: TEST_USER_ID,
       status: 'COMPLETED',
       translationStatus: 'COMPLETED',
       filename: 'original.txt',
@@ -44,10 +50,13 @@ function makeCompletedJobItem(overrides: Record<string, unknown> = {}) {
   );
 }
 
-/** Build a minimal APIGateway event for GET /translation/{jobId}/download */
-const createEvent = (jobId = 'job-123', userId = 'user-abc'): Partial<APIGatewayProxyEvent> => ({
+/** Build a minimal APIGateway event for GET /jobs/{jobId}/download */
+const createEvent = (
+  jobId = TEST_JOB_ID,
+  userId = TEST_USER_ID
+): Partial<APIGatewayProxyEvent> => ({
   httpMethod: 'GET',
-  path: `/translation/${jobId}/download`,
+  path: `/jobs/${jobId}/download`,
   pathParameters: { jobId },
   headers: { origin: 'http://localhost:3000' },
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -58,9 +67,9 @@ const createEvent = (jobId = 'job-123', userId = 'user-abc'): Partial<APIGateway
 });
 
 /** Create an event with no authorizer (simulates unauthenticated request). */
-const createUnauthenticatedEvent = (jobId = 'job-123'): Partial<APIGatewayProxyEvent> => ({
+const createUnauthenticatedEvent = (jobId = TEST_JOB_ID): Partial<APIGatewayProxyEvent> => ({
   httpMethod: 'GET',
-  path: `/translation/${jobId}/download`,
+  path: `/jobs/${jobId}/download`,
   pathParameters: { jobId },
   headers: { origin: 'http://localhost:3000' },
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -83,7 +92,7 @@ describe('downloadTranslation handler', () => {
     dynamoMock.on(GetItemCommand).resolves({ Item: makeCompletedJobItem() });
 
     s3Mock.on(ListObjectsV2Command).resolves({
-      Contents: [{ Key: 'translated/job-123/chunk-0.txt' }],
+      Contents: [{ Key: `translated/${TEST_JOB_ID}/chunk-0.txt` }],
       IsTruncated: false,
     });
 
@@ -111,20 +120,20 @@ describe('downloadTranslation handler', () => {
     // before chunk-2 lexicographically. The handler must sort numerically.
     s3Mock.on(ListObjectsV2Command).resolves({
       Contents: [
-        { Key: 'translated/job-123/chunk-0.txt' },
-        { Key: 'translated/job-123/chunk-10.txt' },
-        { Key: 'translated/job-123/chunk-2.txt' },
+        { Key: `translated/${TEST_JOB_ID}/chunk-0.txt` },
+        { Key: `translated/${TEST_JOB_ID}/chunk-10.txt` },
+        { Key: `translated/${TEST_JOB_ID}/chunk-2.txt` },
       ],
       IsTruncated: false,
     });
 
     // Return different content per key so we can verify ordering.
     s3Mock
-      .on(GetObjectCommand, { Key: 'translated/job-123/chunk-0.txt' })
+      .on(GetObjectCommand, { Key: `translated/${TEST_JOB_ID}/chunk-0.txt` })
       .resolves({ Body: makeS3Stream('chunk0') } as any)
-      .on(GetObjectCommand, { Key: 'translated/job-123/chunk-2.txt' })
+      .on(GetObjectCommand, { Key: `translated/${TEST_JOB_ID}/chunk-2.txt` })
       .resolves({ Body: makeS3Stream('chunk2') } as any)
-      .on(GetObjectCommand, { Key: 'translated/job-123/chunk-10.txt' })
+      .on(GetObjectCommand, { Key: `translated/${TEST_JOB_ID}/chunk-10.txt` })
       .resolves({ Body: makeS3Stream('chunk10') } as any);
 
     const result = await handler(createEvent() as APIGatewayProxyEvent);
@@ -159,7 +168,7 @@ describe('downloadTranslation handler', () => {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       requestContext: {
         requestId: 'req-bad',
-        authorizer: { claims: { sub: 'user-abc' } },
+        authorizer: { claims: { sub: TEST_USER_ID } },
       } as any,
     };
 
@@ -177,24 +186,57 @@ describe('downloadTranslation handler', () => {
   it('returns 404 when job does not exist', async () => {
     dynamoMock.on(GetItemCommand).resolves({ Item: undefined });
 
-    const result = await handler(createEvent('missing-job') as APIGatewayProxyEvent);
+    // Use a valid UUID that doesn't match any seeded job.
+    const result = await handler(
+      createEvent('99999999-9999-9999-9999-999999999999') as APIGatewayProxyEvent
+    );
 
     expect(result.statusCode).toBe(404);
   });
 
-  it('returns 404 when job belongs to a different user (BOLA prevention)', async () => {
-    // The DynamoDB composite key query for the attacker's userId returns
-    // no Item — the handler cannot distinguish "not found" from
-    // "belongs to someone else", so both produce 404.
-    dynamoMock.on(GetItemCommand).resolves({ Item: undefined });
+  it('returns 404 when job belongs to a different user (BOLA prevention) — non-vacuous', async () => {
+    // Non-vacuous ownership test: the DDB mock returns the real job ONLY when
+    // queried with the owner's composite key (jobId + ownerUserId). A request from
+    // attackerUserId queries a different composite key and gets Item: undefined.
+    //
+    // This test FAILS if you remove the ownership check from the handler because
+    // without it the handler would fetch the job item unconditionally (or skip
+    // the DDB call entirely), returning 200 instead of 404.
+    // Must be valid UUIDs so the UUID-format guard doesn't reject early.
+    const ownerUserId = 'aaaaaaaa-0000-0000-0000-000000000001';
+    const attackerUserId = 'bbbbbbbb-0000-0000-0000-000000000002';
+    const jobId = 'cccccccc-0000-0000-0000-000000000003';
 
-    const result = await handler(createEvent('job-123', 'attacker-xyz') as APIGatewayProxyEvent);
+    const ownerKey = marshall({ jobId, userId: ownerUserId });
+    const attackerKey = marshall({ jobId, userId: attackerUserId });
 
-    expect(result.statusCode).toBe(404);
-    // Response body must not contain the job's actual userId or jobId data
-    const body = JSON.parse(result.body);
-    expect(body).not.toHaveProperty('userId');
-    expect(body).not.toHaveProperty('translationStatus');
+    // Seed: owner's key → real job record; attacker's key → not found.
+    dynamoMock
+      .on(GetItemCommand, { Key: ownerKey })
+      .resolves({ Item: makeCompletedJobItem({ jobId, userId: ownerUserId }) })
+      .on(GetItemCommand, { Key: attackerKey })
+      .resolves({ Item: undefined });
+
+    // Attacker's request — must get 404, not the job content.
+    const attackerResult = await handler(
+      createEvent(jobId, attackerUserId) as APIGatewayProxyEvent
+    );
+    expect(attackerResult.statusCode).toBe(404);
+    const attackerBody = JSON.parse(attackerResult.body);
+    expect(attackerBody).not.toHaveProperty('userId');
+    expect(attackerBody).not.toHaveProperty('translationStatus');
+
+    // Sanity-check: owner's request returns 200 (proves the mock is set up correctly).
+    // Without this, a mock that ALWAYS returns undefined would also pass the attacker test,
+    // making it a vacuous green.
+    s3Mock.on(ListObjectsV2Command).resolves({
+      Contents: [{ Key: `translated/${jobId}/chunk-0.txt` }],
+      IsTruncated: false,
+    });
+    s3Mock.on(GetObjectCommand).resolves({ Body: makeS3Stream('content') } as any);
+
+    const ownerResult = await handler(createEvent(jobId, ownerUserId) as APIGatewayProxyEvent);
+    expect(ownerResult.statusCode).toBe(200);
   });
 
   // -------------------------------------------------------------------------
@@ -262,7 +304,7 @@ describe('downloadTranslation handler', () => {
     dynamoMock.on(GetItemCommand).resolves({ Item: makeCompletedJobItem() });
 
     s3Mock.on(ListObjectsV2Command).resolves({
-      Contents: [{ Key: 'translated/job-123/chunk-0.txt' }],
+      Contents: [{ Key: `translated/${TEST_JOB_ID}/chunk-0.txt` }],
       IsTruncated: false,
     });
 
@@ -297,7 +339,7 @@ describe('downloadTranslation handler', () => {
     });
 
     s3Mock.on(ListObjectsV2Command).resolves({
-      Contents: [{ Key: 'translated/job-123/chunk-0.txt' }],
+      Contents: [{ Key: `translated/${TEST_JOB_ID}/chunk-0.txt` }],
       IsTruncated: false,
     });
 
@@ -316,7 +358,7 @@ describe('downloadTranslation handler', () => {
     });
 
     s3Mock.on(ListObjectsV2Command).resolves({
-      Contents: [{ Key: 'translated/job-123/chunk-0.txt' }],
+      Contents: [{ Key: `translated/${TEST_JOB_ID}/chunk-0.txt` }],
       IsTruncated: false,
     });
 
@@ -336,7 +378,7 @@ describe('downloadTranslation handler', () => {
     dynamoMock.on(GetItemCommand).resolves({ Item: makeCompletedJobItem() });
 
     s3Mock.on(ListObjectsV2Command).resolves({
-      Contents: [{ Key: 'translated/job-123/chunk-0.txt' }],
+      Contents: [{ Key: `translated/${TEST_JOB_ID}/chunk-0.txt` }],
       IsTruncated: false,
     });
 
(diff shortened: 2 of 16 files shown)


leixiaoyu:
## Performance Review (OMC) — PR #212

**Critical**: None — for the 65K–400K-word demo target the design fits within Lambda + API Gateway limits. Parallel `Promise.all` fetch (`downloadTranslation.ts:223`), Node 22 + arm64, 512 MB / 60 s, and read-only IAM are all reasonable choices for the POC.

**Recommended**:
- **API Gateway 10 MB hard limit (`downloadTranslation.ts:259`)**: At 400K words, plain ASCII English ≈ 2.5 MB; CJK / multi-byte UTF-8 expansion can push this 2–4×. Combined with `'Cache-Control': 'no-store'` (no compression negotiation guaranteed at API GW), there is real headroom risk on worst-case docs. Recommend either (a) a hard guard `if (assembledDocument.length > 6 * 1024 * 1024) return 413` with a clear message, or (b) for the next iteration, return a presigned S3 URL to a freshly-assembled object. Both avoid a confusing API GW `BodySizeTooLarge` 500 in front of a demo audience.
- **Unbounded `Promise.all` fan-out (`downloadTranslation.ts:223`)**: 150 concurrent `GetObject`s on a single Lambda is fine for S3, but the AWS SDK v3 default `NodeHttpHandler` has no explicit `keepAlive` agent or `maxSockets` tuning here. For the worst-case 150-chunk doc you'll thrash TCP connections. Add `requestHandler: new NodeHttpHandler({ httpsAgent: new https.Agent({ keepAlive: true, maxSockets: 50 }) })` to the `S3Client`, or chunk into batches of ~25 with `p-limit`. Easy 2–4× wall-time win on the cold path.
- **Memory headroom for assembly (`downloadTranslation.ts:227`)**: `chunkContents.join('\n')` for ~150 strings of ~10 KB each is fine at 512 MB, but each chunk currently lives as a separate string AND in the joined string simultaneously (peak ≈ 2× total bytes ≈ 5 MB worst case). Safe for the demo target, but watch this if document size grows. Consider streaming via `Readable.from(chunkContents)` if you ever raise the cap above 400K words.

**Suggestions**:
- Emit a CloudWatch metric for `chunkCount`, `documentBytes`, and end-to-end duration on the success path (`downloadTranslation.ts:229`) so you can set a perf budget alert (e.g., p95 < 3 s for ≤ 150 chunks).
- The download endpoint is invoked once per (job, language) pair. For the multi-language demo (5 target langs from one source), each request triggers an independent `ListObjectsV2` + 150 `GetObject`s. Acceptable for the demo, but if the user re-downloads, consider caching the assembled blob to S3 (`assembled/{jobId}.txt`) on first download — turns subsequent downloads into a single redirect / passthrough.
- `Cache-Control: no-store` (`downloadTranslation.ts:257`) is correct for security but blocks any CDN/browser caching. Worth revisiting once content is final.

**Positive notes**:
- Parallel fetch via `Promise.all` (`downloadTranslation.ts:223`) — the right call vs. sequential.
- Numeric chunk sort (`downloadTranslation.ts:104`) avoids a re-fetch correctness bug.
- `documentBytes` logged as length only, never content (`downloadTranslation.ts:233`) — no log-volume blowup, no PII risk on the hot path.
- arm64 + Node 22 + 512 MB matches existing translateChunk profile — predictable cold start (~400–800 ms).
- Pagination on `ListObjectsV2` (`downloadTranslation.ts:84-100`) handles >1000 chunks correctly even though current docs won't hit that.
- Module-scoped `s3Client` / `dynamoClient` (`downloadTranslation.ts:53-54`) — connection reuse across warm invocations is preserved.

leixiaoyu:
## Architecture Review (OMC) — PR #212

**Architectural impact: MEDIUM** — one new public resource path that breaks the established URL convention, an incomplete SSoT migration, and a stack file that crossed an LOC threshold where modularization stops being optional.

---

**Critical**:

- **URL convention drift — `/translation/{jobId}/download` vs the established `/jobs/{jobId}/...` pattern.** Every other job-scoped translation operation lives under `/jobs/{jobId}/...`: `/jobs/{jobId}/translate` (`lfmt-infrastructure-stack.ts:1786`), `/jobs/{jobId}/translation-status` (`:1813`), `/jobs/{jobId}/progress` (`:517`), `/jobs/{jobId}/status` (`:518`). This PR introduces a parallel `/translation` root resource (`:640`) solely because `translationService.ts:347` was written that way first. The PR body is honest about this (\"to match the frontend's URL exactly without requiring a service-layer change\") but the architecturally correct fix is one frontend-line change to `/jobs/${jobId}/download`, not a permanent second URL hierarchy. Two consequences if shipped as-is: (a) every future translation operation will face a coin-flip about which root to live under; (b) the new `/translation/{jobId}` resource is a degenerate hierarchy with exactly one child (`download`), which is a code-smell pointing at the misplacement. **Recommendation**: change the route to `/jobs/{jobId}/download`, change the one frontend line, delete the new resource tree.

- **SSoT for `TranslationTone` is incomplete on the consumer that matters most.** `frontend/src/components/Translation/TranslationConfig.tsx:47` still defines `TONE_OPTIONS` as a local literal array — it does not import `TRANSLATION_TONE_VALUES` from `@lfmt/shared-types`. The `toneContract.test.ts` even hardcodes a duplicated `FRONTEND_TONE_VALUES` and admits in a comment that it is not actually importing the frontend constant. Result: the SSoT covers backend-to-backend drift but the original failure mode (frontend offers a tone the backend rejects) is still possible — TypeScript will not catch a new entry added to `TONE_OPTIONS` if it is not also added to `TRANSLATION_TONE_VALUES`. **Recommendation**: rewrite `TONE_OPTIONS` as `TRANSLATION_TONE_VALUES.map(value => ({ value, label: TONE_LABELS[value], description: TONE_DESCRIPTIONS[value] }))` so the union literally is the source.

**Recommended**:

- **`TranslationTone` is in the wrong shared-types module.** It lives in `shared-types/src/jobs.ts` (`:18`) but tone is a *translation parameter*, not a job-state field. Sibling modules already exist by domain (`auth.ts`, `documents.ts`, `legal.ts`, `polling.ts`, `workflows.ts`); a `translation.ts` module would be the natural home and would also house `DownloadTranslationApiResponse` (currently also misplaced in `jobs.ts:344`). The current placement makes `jobs.ts` accumulate orthogonal concerns and makes \"where does this belong?\" harder for the next contributor. Low cost to relocate now, high cost once 5+ callers import from the wrong path.

- **CDK stack now 2,242 LOC** (`lfmt-infrastructure-stack.ts`). Issue #64 (nested stacks) was deferred when the file was ~1,500 LOC; this PR adds another Lambda + dedicated role + 4 API Gateway resources + CORS preflights for each. The pattern of \"one new endpoint = ~150 LOC scattered across constructor, role-creation block, lambda-creation block, API-routing block\" is now unmistakable. **Recommendation**: bump #64 priority — at minimum extract a `TranslationApiConstruct` and `TranslationLambdaConstruct` so future endpoints become ~30 LOC each.

- **\"Scoped job-related Lambda role\" pattern is now 3+ duplicated blocks.** `deleteJobRole` (PR #208), now `downloadTranslationRole` (`:551`), and the older `translationRole` all follow the same shape: ServicePrincipal + AWSLambdaBasicExecutionRole managed policy + a ManagedPolicy with bespoke statements. Worth a small helper (`createScopedJobLambdaRole({ id, description, statements })`) or a CDK construct. Three is the canonical \"extract\" threshold and the next two endpoints (cancel-job, retry-job, etc.) will repeat it.

**Suggestions**:

- **Multi-chunk assembly is lazy at download time, in-memory.** The handler does `Promise.all(validChunkKeys.map(fetchChunkContent))` then `chunkContents.join('\\n')` (`downloadTranslation.ts:634-638`). For a 400K-word doc (~115 chunks, ~3 MB raw), this fits in 512 MB Lambda memory and under the 6 MB API Gateway response limit — but the architectural choice is implicit, not documented. Two failure modes the architecture should acknowledge: (a) at ~2× current chunk-size, response payload exceeds the 6 MB sync limit and the endpoint silently breaks; (b) every download recomputes assembly — wasteful for large jobs that get re-downloaded. **Suggestion**: either materialize the assembled doc once at workflow completion (Step Functions terminal step writes `translated/{jobId}/final.txt`) and have the download Lambda just stream that single object via presigned URL, or document the 6 MB ceiling as a known constraint with a tracked issue. The presigned-URL pattern also collapses the dedicated Lambda+role to almost nothing.

- **No streaming for large responses.** Lambda response streaming (`awslambda.streamifyResponse`) is supported and would lift the 6 MB ceiling toward 20 MB. Architecturally consistent with the \"large-document translation service\" positioning. Worth a tracked issue even if not done in this PR.

- **Smoke fixture location.** `frontend/e2e/fixtures/smoke-test-minimal.txt` couples a backend invariant (`MIN_FILE_SIZE = 1000` from `shared-types/src/validation.ts`) to a frontend test asset. If the backend ever raises the floor, the fixture goes stale silently. **Suggestion**: either centralize fixtures under a top-level `test-fixtures/` (consumed by both frontend e2e and backend integration tests) or add a backend test that asserts the fixture is above `MIN_FILE_SIZE` so the coupling is enforced, not implicit.

- **Boy-scout: PascalCase vs kebab-case CDK logical IDs.** `infrastructure.test.ts:486-492` searches the synthesized template for `'DownloadTranslation'` (PascalCase) in the integration URI. Several previous PRs hit this exact gotcha because CDK normalizes IDs differently than function names. A small helper `expectLambdaIntegration({ template, idPrefix, httpMethod })` would prevent the next test author from re-discovering it.

**Positive notes**:

- **Dedicated IAM role + read-only scoping is exemplary.** `downloadTranslationRole` grants only `dynamodb:GetItem`, `s3:GetObject` on `translated/*`, and `s3:ListBucket` (`:551-583`) — strictly less than `translationRole`, and the drift-guard tests assert the *absence* of `s3:PutObject` / `s3:DeleteObject` (`infrastructure.test.ts:412-433`). This is the right pattern; the recommendation above is to extract it, not to abandon it.

- **BOLA prevention via composite-key lookup is architecturally clean.** `loadJobForUser()` returning `null` for both \"missing\" and \"wrong owner\" makes the 404 collapse at the data-access layer, not at the handler — much less likely to leak via a future code path. The test that asserts the error body contains no `userId` / `translationStatus` (`downloadTranslation.test.ts:241-243`) is the right shape.

- **409 vs 404 for non-COMPLETED jobs is the architecturally honest choice.** Surfaces the lifecycle state to the frontend instead of pretending the resource does not exist. Aligns with the broader system's explicit-state-machine design.

- **Drift-guard headcount discipline preserved.** Bumping `EXPECTED_APPLICATION_LAMBDA_COUNT` from 13 → 14 with a comment trail explaining why is the right pattern and matches the precedent set by PR #208.

leixiaoyu:
## Security Review (OMC) — PR #212

Lane: security. Scope: new `GET /translation/{jobId}/download` endpoint that exposes translated user content; tone SSoT; smoke fixture.

**Critical**
- None. The auth model on the new endpoint is sound: Cognito authorizer enforced at API GW (`lfmt-infrastructure-stack.ts:1942-1948`), `userId = authorizer.claims.sub` (`downloadTranslation.ts:564`), and ownership via composite-key `loadJobForUser` (`downloadTranslation.ts:579`). Cross-user access returns 404 (`:582`), preventing existence-leak (OWASP API1:2023 — BOLA). Status guard returns 409, not 200, for non-COMPLETED jobs (`:590-598`). Dedicated IAM role with read-only S3 scoped to `translated/*` (`lfmt-infrastructure-stack.ts:1031-1055`). Drift guards added (`infrastructure.test.ts:854-980`).

**Recommended**
- **(R1) `Content-Disposition` is interpolated, not encoded** (`downloadTranslation.ts:665`): `filename="${downloadFilename}"`. Currently SAFE only because `filenameSchema` (`shared-types/src/validation.ts:22`) restricts uploaded filenames to `^[a-zA-Z0-9._-]+\.txt$` — no quotes, semicolons, or CRLF can land in `job.filename`. This is a transitive-dep on a separate file's regex; if that schema is ever loosened to accept spaces/Unicode (very plausible for i18n), this becomes a header-injection / `Content-Disposition` parsing issue. Defense-in-depth: pass the value through `encodeURIComponent` for the RFC 6266 `filename*=UTF-8''…` form, or strip `["\r\n;\\]` at this seam. Add a unit test that asserts `filename="..."` is preserved when `job.filename` contains an unexpected character (proving the sanitiser fires).
- **(R2) Missing-chunk-in-the-middle case silently produces partial output** (`downloadTranslation.ts:603-638`). The handler only fails when `chunkKeys.length === 0` or all keys have unparseable indices. If S3 returns chunks `[0,1,3]` (chunk 2 lost / not yet written / racing with cleanup), the user receives a corrupt document with no warning, no log, and HTTP 200. Suggest: compute `expected = max(idx) + 1`, compare to `validChunkKeys.length`, return 500 with `requestId` (and log `expected vs got`) when they disagree. This is a data-integrity guard, not a security bug, but matters for "did we leak a partial document the user can't tell is partial."
- **(R3) Memory-load model is unbounded** (`downloadTranslation.ts:634-638`). `Promise.all` + `join('\n')` materialises all chunks in memory. PR comment claims "115 chunks at 3500 tokens" for 400K words — at ~5 bytes/token that's ~2 MB string post-translate, ~4 MB UTF-16 in V8. Within the 512 MB Lambda budget but the next size class (1M-word legal contracts which the spec mentions) would push past safe headroom and the failure mode is OOM → 500 → user retries → repeated S3 GET billing. Cleanest medium-term fix is the presigned-URL response pattern (frontend GETs a single concatenated S3 object, or the lambda returns N presigned URLs and the frontend stitches in the browser). Acceptable for the demo at current sizes; flag for post-demo follow-up. (Performance-engineer is the primary lane here; flagging because the OOM mode is also a DoS-via-large-file vector.)
- **(R4) Logging enumerates `chunkKeys` on the corrupt-keys error path** (`downloadTranslation.ts:621`). Keys are `translated/{jobId}/chunk-N.txt` — not PII, but logging the full list on every corrupt-job request creates a high-volume log surface for an attacker who can trigger the corrupt-key path. Cap at first 5 + count. (Verified on the success path the handler logs only `chunkCount` and `documentBytes`, not content — good.)

**Suggestions**
- **(S1) UUID-format check on `jobId` path param** (`downloadTranslation.ts:570-573`). `jobId` is currently any non-empty string; `loadJobForUser` will return null for nonsense values which is correct, but a `z.string().uuid()` check at the boundary returns 400 instead of consuming a DDB RCU per probe. Same hardening applies to GetJob / DeleteJob (existing surface) — if you decide to do it, do it in `loadJobForUser` once, not three places.
- **(S2) `Cache-Control: no-store` on download** (`downloadTranslation.ts:668`) is correct for the translated payload. Consider also `X-Content-Type-Options: nosniff` and `Referrer-Policy: no-referrer` on this response specifically — translated content can contain attacker-influenced bytes (the user uploaded the source) and a strict-MIME header costs nothing.
- **(S3) `s3:ListBucket` is bucket-wide** (`lfmt-infrastructure-stack.ts:1049-1053`). The PR comment correctly notes that prefix-scoping `ListBucket` requires an IAM `Condition` block (`s3:prefix` with `StringLike: ["translated/*"]`) — without it, the role can list `uploads/`, `chunks/`, `documents/`, etc. (read metadata: keys + sizes + timestamps). It cannot read those objects (GetObject is properly prefix-scoped), so impact is limited to bucket-listing reconnaissance, but the agent comment claims this is intentional — the standard hardening is the prefix `Condition`. Worth doing.
- **(S4) Tone SSoT** (`shared-types/src/jobs.ts:33-46`, `startTranslation.ts:33`): Validator now uses `TRANSLATION_TONE_VALUES`. No injection vector — values are an `as const` literal array, runtime check is `Array.includes`. `toneContract.test.ts:75-87` pins the canonical set. No security implications. (Minor: the contract test re-declares `BACKEND_VALIDATION_TONE_VALUES` as an inline literal — if a future refactor renames `TRANSLATION_TONE_VALUES` the test may pass while real backend validation drifts. Consider importing the validator or asserting on its exported allowlist directly.)
- **(S5) Smoke fixture** (`smoke-test-minimal.txt:19-24`): padding text is generic prose about translation theory. No PII, no credentials, no secret-shaped data. Safe.

**Positive notes**
- Dedicated `DownloadTranslationLambdaRole` (`lfmt-infrastructure-stack.ts:1023-1055`) is correctly isolated from the shared `translationRole` (which has `PutObject`/`DeleteObject`). Drift guard `DownloadTranslationLambdaRole does NOT have s3:PutObject or s3:DeleteObject` (`infrastructure.test.ts:884-905`) is exactly the test I would have asked for.
- 404 on cross-user (`downloadTranslation.test.ts:231-244`) explicitly asserts response body does not contain `userId` or `translationStatus` — the right shape for "no information leakage on auth failure."
- 409 with current status (`downloadTranslation.ts:592-597`, `downloadTranslation.test.ts:250-285`) is the right code: resource exists, state forbids the action. Better than 404 (which would conflate with BOLA).
- `EXPECTED_APPLICATION_LAMBDA_COUNT` 13 → 14 (`infrastructure.test.ts:841`) keeps the headcount drift guard honest. Same pattern as PR #208.
- Path-traversal via `jobId` is mitigated by the DDB composite-key lookup happening BEFORE the S3 prefix interpolation: any attacker-crafted jobId that does not exactly match a row the attacker owns returns 404 at line 582, never reaching the S3 call. Good ordering.
- No translated content logged at any level (`downloadTranslation.ts` — info-level logs only `requestId`, `jobId`, `chunkCount`, `documentBytes`). Correct PII posture.
- `responseType: 'blob'` flow on the frontend means the response body is never interpolated into the DOM — no XSS surface even though translated text could contain `<script>` from a malicious uploaded source.

References: OWASP API1:2023 (BOLA), OWASP API3:2023 (Broken Object Property Level Authorization), RFC 6266 (Content-Disposition).

leixiaoyu:
## Test Coverage Review (OMC) — PR #212

**Critical**:

1. **BOLA test verifies the wrong mechanism** (). The test description says "the DynamoDB composite key query for the attacker's userId returns no Item", but the mock is `dynamoMock.on(GetItemCommand).resolves({ Item: undefined })` — unconditional, so it would return the same 404 whether the handler checks ownership or not. The test does not prove that `loadJobForUser` enforces the composite key; a handler that only does a `GetItem` on jobId alone (ignoring userId) would pass this test identically. A correct BOLA test must mock the victim's item as found when queried with the victim's userId but missing when queried with the attacker's userId (e.g., using `mockClient.on(GetItemCommand, { Key: { jobId: ..., userId: 'attacker-xyz' } })` vs a separate matcher for the victim). As written, the BOLA coverage is vacuous.

2. **`toneContract.test.ts` does not actually cross-check the backend validator** (). The test defines `BACKEND_VALIDATION_TONE_VALUES` as a hand-typed local constant and compares it to `TRANSLATION_TONE_VALUES`. That detects nothing: if a developer adds a fourth tone to `TRANSLATION_TONE_VALUES` but forgets to update `startTranslation.ts` line 252 (the inline `includes()` guard), the contract test still passes because both it and the shared-types array would agree. The test would need to parse or import the actual runtime validator from `startTranslation.ts` to catch that drift. Additionally, the PR diff shows `startTranslation.ts` was changed to use `TRANSLATION_TONE_VALUES` — but inspecting the file confirms the inline array at line 252 remains unchanged (`!['formal', 'informal', 'neutral'].includes(body.tone)`). The contract test cannot detect this divergence.

3. **E2E error-scenario intercept uses the wrong URL pattern** (). The route intercept is `**/v1/translation/jobs//download` but the actual endpoint (and MSW handler, and translationService.ts line 347) is `/translation//download` (no `/jobs/` segment). The intercept will never fire; the test silently skips the assertion via the surrounding `try/catch`.

**Recommended**:

4. **Multi-chunk ordering test is present and correct** (), but the numeric spread only covers `chunk-0 / chunk-2 / chunk-10`. Adding a case where S3 returns keys with a gap that also exposes NaN-sort behaviour (e.g., an unrecognised filename in the prefix) would exercise the `validChunkKeys` filter path (), which currently has no dedicated test.

5. **Infra test IAM assertion is role-description-gated, not policy-content-gated** (). The role lookup uses `role.Properties?.Description?.includes('download-translation')`. If the role description changes or is omitted, the policy assertions at lines 697-747 silently become no-ops because `downloadRoleEntry` would be `undefined` and the test falls through to `expect(downloadRoleEntry).toBeDefined()` (which does catch it), but the `s3:GetObject` scope check at line 738 would never run. Consider asserting the logical-ID match (e.g., `DownloadTranslationLambdaRole` PascalCase prefix) as a secondary guard.

6. **Frontend `ToneCode` is not linked to `TranslationTone` from shared-types** (). `ToneCode` is derived from the local `TONE_OPTIONS` array, not from `TRANSLATION_TONE_VALUES`. The compiler therefore cannot catch drift between the frontend selector and the backend type. The contract test comment acknowledges frontend parity is delegated to `TranslationConfig.test.tsx "should show all 3 tone options"` — but that test only verifies UI rendering, not type-level alignment with `TranslationTone`.

7. **`translationService.test.ts` has only one success scenario and one error scenario for `downloadTranslation`** (lines 748-796). Missing: 409 (job not yet complete) — the frontend service calls the same endpoint and would receive a JSON error body from a Lambda whose 200 path returns `text/plain`, so the axios Blob deserialization for error responses deserves a test.

**Suggestions**:

8. Empty S3 chunk (zero-byte object exists) is not tested. `fetchChunkContent` calls `transformToString` and only guards against a missing `Body` (); a zero-byte body would succeed and inject an empty string silently into the assembled document.

9. **Smoke fixture growth** (): no tests assert a content hash or specific byte offset. The fixture comment explains the 1,000-byte minimum requirement — confirming the new size (1,354 bytes) does not exceed any maximum checked by `fileSizeSchema` would be valuable to document in the file header or a comment in the validator, but no regression risk is introduced by the growth itself.

10. The infra test asserts `DownloadTranslationFunction` runtime and architecture () — this directly mirrors the PR #203 drift guard pattern and is correct. The COGNITO authorizer assertion at line 770 is also present and well-scoped.

**Positive notes**:

- All 7 originally scoped unit test cases are present; the multi-chunk numeric-vs-lexicographic ordering test () is explicit and would catch a naive `.sort()` regression.
- The CORS header tests on both success (line 182) and error (line 198) paths close the cross-origin download regression gap.
- The `full-flow-mock.spec.ts` E2E test (line 199-207) validates the complete journey through the MSW handler end-to-end, including content assertion (`[MOCK TRANSLATION COMPLETE]` marker) and the no-AWS-hosts guard — this is the strongest end-to-end coverage in the PR.
- Infra test correctly uses PascalCase `DownloadTranslation` in URI matching (line 765), avoiding the logical-ID case-mismatch that was fixed from previous PRs.
- Dedicated `downloadTranslationRole` negative assertions (lines 685-705) for `s3:PutObject` / `s3:DeleteObject` directly protect against privilege-creep regression.

leixiaoyu:
## Code Quality Review (OMC) — PR #212

**Critical** (must fix):
- None.

**Recommended**:

1. **`downloadTranslation.ts` duplicates the project's shared `createErrorResponse` helper** (`backend/functions/shared/api-response.ts`). Every other Lambda in this codebase — `getJob.ts`, `deleteJob.ts`, `uploadRequest.ts` — calls `createErrorResponse(statusCode, message, requestId, undefined, requestOrigin)`. This Lambda defines its own private `errorResponse()` with a subtly different signature (`requestId` always included in body, `requestOrigin` is last not middle). The justification in the doc comment (raw text/plain 200 response requires different headers) is valid for the *success path* only; the error shape is identical JSON. Switching error branches to `createErrorResponse` removes the helper duplication and keeps the response contract uniform. File: `backend/functions/translation/downloadTranslation.ts:281-306`.

2. **`toneContract.test.ts` comment on `BACKEND_VALIDATION_TONE_VALUES` is already stale** at merge time. The comment at `toneContract.test.ts:46-47` says the backend uses "its inline includes() guard" and references `startTranslation.ts:252` — but this same PR replaced that inline array with `TRANSLATION_TONE_VALUES` (commit `47de621`). The constant `BACKEND_VALIDATION_TONE_VALUES` is now a manual duplicate of the SoT, not a snapshot of an inline literal. Either: (a) delete the constant and instead assert `TRANSLATION_TONE_VALUES` directly (the validator already imports it), or (b) update the comment to be accurate. File: `backend/functions/translation/toneContract.test.ts:44-51`.

3. **`TRANSLATION_TONE_VALUES` uses `ReadonlyArray<TranslationTone>` annotation rather than `as const satisfies`** — the same pattern already used for `TRANSLATION_TERMINAL_STATUSES` and `CHUNKING_ERROR_STATUSES` three blocks above in the same file (`shared-types/src/jobs.ts:57-75`). With the current annotation the array literal widens to `ReadonlyArray<TranslationTone>`, so narrowing to literal element types requires a cast at the call-site in `startTranslation.ts` (`(TRANSLATION_TONE_VALUES as ReadonlyArray<string>).includes(body.tone)`). Using `as const satisfies ReadonlyArray<TranslationTone>` would preserve literal narrowing and match the existing pattern in the same file, removing the need for the cast. File: `shared-types/src/jobs.ts:30-34`.

4. **CORS `defaultCorsPreflightOptions` block copy-pasted three times** for `translationRootResource`, `translationJobResource`, and `downloadResource` (`lfmt-infrastructure-stack.ts:1933-1981`). The existing `createApiGateway()` method sets `defaultCorsPreflightOptions` at the REST API root level, which CDK normally inherits. The three repeated blocks are identical except for `allowMethods: ['GET', 'OPTIONS']` vs the root's `['GET', 'POST', 'PUT', 'DELETE', 'OPTIONS']`. If GET+OPTIONS is intentional for this path subtree, extract a `translationCorsPreflight` constant before the first `addResource` call. If the narrower methods list is not required, omit `defaultCorsPreflightOptions` on child resources and let the root-level default propagate.

5. **No test for multi-page S3 `ListObjectsV2` pagination.** `listTranslatedChunkKeys` correctly loops on `IsTruncated`, but every test in `downloadTranslation.test.ts` sets `IsTruncated: false` on a single mock response. A 400K-word document at 3,500 tokens/chunk produces ~115 chunks; S3 `ListObjectsV2` pages at 1,000 objects by default so pagination is not exercised in practice, but a test with two pages (e.g. chunk-0..chunk-5 on page 1, chunk-6..chunk-10 on page 2) would pin the loop logic and document the numeric-sort-after-pagination contract. File: `backend/functions/translation/downloadTranslation.test.ts`.

**Suggestions**:

- **`DownloadTranslationApiResponse` interface is marked "documentation only / not currently used"** but is exported from `shared-types/src/jobs.ts`. Unused exports accumulate over time and can mislead future contributors into believing the endpoint returns JSON. If it is genuinely only documentation, a `// @internal` JSDoc tag or moving it to a code comment inside the Lambda file would make the intent clearer than exporting a live type.

- **Boy-scouts: add a one-line comment in `infrastructure.test.ts`** near the `DownloadTranslation` API Gateway method test (`infrastructure.test.ts:1743-1744`) explaining why the URI search uses `DownloadTranslation` (PascalCase, CDK-generated logical ID) rather than `download-translation`. The agent note already identifies that the kebab vs PascalCase distinction cost a search iteration; a single sentence here will save the next contributor the same investigation.

- The `s3Key` field referenced by `deleteJob.ts` is not present in `DynamoDBJob` in `shared-types` (`s3Key` is not typed on the interface, only `[key: string]: unknown` catches it). The download Lambda correctly avoids relying on `s3Key`, but the gap in the shared type is unrelated pre-existing debt worth tracking.

**Positive notes**:

- Numeric sort implementation in `parseChunkIndex` + `listTranslatedChunkKeys` is clean: the `NaN` guard (`isNaN()` filter on `validChunkKeys`) closes the edge case where an unrecognised key format would silently corrupt sort order. Well-documented with inline comments.
- `createJobLambda()` helper is used correctly for `DownloadTranslationFunction` — no inlining of the `NodejsFunction` block.
- IAM role isolation is correctly scoped: `s3:GetObject` limited to `translated/*` prefix, no `PutObject`/`DeleteObject`. The infrastructure test `DownloadTranslationLambdaRole does NOT have s3:PutObject or s3:DeleteObject` pins this contract.
- `requestId` threading follows the project standard throughout: extracted from `event.requestContext.requestId` and threaded into every log call and every error response body.
- Smoke fixture content is semantically appropriate for a translation test and the explicit byte count comment (`MIN_FILE_SIZE = 1000 bytes`) makes the intent of the padding immediately clear to the next reader.

xlei-raymond (approved):
Detailed Review Findings

  1. Structural Integrity & Handshake (The Download Fix)
  The PR correctly implements the high-priority GET /jobs/{jobId}/download endpoint.
   * The Fix: A sophisticated Lambda handler that performs parallel chunk assembly using ListObjectsV2 and Promise.all. The sorting logic (numerical index extraction) correctly handles the "chunk-10 vs chunk-2"
     lexicographical pitfall.
   * Safety: The use of a dedicated DownloadTranslationLambdaRole (Role 6) with prefix-scoped s3:GetObject and s3:ListBucket permissions is a perfect implementation of least-privilege.
   * Handshake: The inclusion of the 409 Conflict check for non-COMPLETED jobs ensures that the UI cannot attempt to download a half-baked translation.

  2. Architectural Excellence (TranslationTone SSoT)
  The most significant win in this PR is the centralization of TranslationTone.
   * The Problem: The allowed set of tones ('formal', 'informal', 'neutral') was duplicated across five different source files, creating a high risk of "silent drift" if a new tone were added.
   * The Solution: Moving the type and values into shared-types/src/jobs.ts and wiring them through all layers (Lambda validation, Component state, Page logic).
   * Unit-Speed Regression Guard: The two new contract tests in TranslationConfig.test.tsx and the toneContract.test.ts on the backend are exceptional. They ensure the frontend and backend stay in sync with the shared
     type in milliseconds.

  3. Resilience & Best Practices (OMC follow-ups)
   * Performance (Rec 15): The use of NodeHttpHandler with keep-alive and maxSockets: 50 is a high-standard optimization for the chunk-assembly phase.
   * BOLA Verification (Critical 1): I highly commend the non-vacuous BOLA test. By mocking a mismatched userId and asserting a 404, the team has provided actual proof that the download endpoint respects ownership.
   * Content-Disposition (Rec 10): The sanitization of the filename via SAFE_FILENAME_PATTERN is an excellent defensive measure against header-injection vulnerabilities.

  4. Verification Results
   * Unit Tests: All 21 new download tests and the updated TranslationConfig tests pass perfectly.
   * Contract Integrity: Verified the 1:1 mapping between TRANSLATION_TONE_VALUES and the UI dropdown.
   * DX Fix: The update to the smoke fixture ensures that our verified post-deploy signal is no longer blocked by trivial size violations.

   * Act Warnings: As with previous frontend PRs, I observed several act(...) warnings in the test output. These should be addressed in the upcoming "Test Hygiene" sweep but are not blockers for this critical demo PR.
   * Response Sizing (Rec 14): The 6 MB body-size guard is a prudent response to API Gateway limits. I agree with the decision to surface a 413 and point to future presigned-URL enhancements for extremely large books.

  Verdict: APPROVE (GO)
  This PR is a definitive upgrade to the LFMT project's production readiness. It resolves the last high-priority demo blocker and establishes robust, self-documenting patterns for our system's most sensitive data paths.

  ---

  PR Review for GitHub

  Status: ✅ Approved

  This PR represents an exceptional maturation of our system's data-retrieval architecture and a major win for operational stability. By implementing the missing Download endpoint with strict least-privilege guards and
  centralizing our TranslationTone Single Source of Truth, the team has successfully cleared the final high-priority blockers for the Investor Demo.

   1. Robust Parallel Chunk Assembly: The implementation of the download handler correctly addresses the complexity of re-assembling 400K-word documents. The combination of parallel GetObject calls and deterministic
      numerical sorting ensures that the assembled document preserves the original's logical flow without the latency of sequential fetching.
   2. Gold-Standard Least Privilege (Role 6): I highly commend the creation of the isolated DownloadTranslationLambdaRole. Restricting the handler to read-only access on the translated/* prefix is a textbook display of
      defensive infrastructure engineering, protecting our source documents from accidental disclosure through the assembly service.
   3. Architectural Alignment (SSoT): Consolidating our translation tone definitions in shared-types and deriving our UI metadata and backend validation from these constants is a major win for system integrity. It
      effectively eliminates the "configuration drift" bug class for our translation settings.
   4. Resilient Content Disposition: The use of a sanitized filename with a safe-fallback policy in the Content-Disposition header demonstrates high attention to detail and a commitment to robust browser-level user
      experience.

  Final Verification Results
   * Logic Integrity: 100% (BOLA guards and assembly ordering verified).
   * Regression Guard: 100% (SSoT contract and Role isolation verified).
   * CI Impact: Smoke and Download gates unblocked.

  ---
  Status: 100% Logic Verified; Security Surface Hardened; Architectural Gaps Closed.

Issue #236, opened 2026-05-10
TranslationDetail: 403 error handler calls setTimeout in render path


leixiaoyu:
## Finding

**File:** `frontend/src/pages/TranslationDetail.tsx`
**Severity:** Medium
**Detected in:** OMC R1 for #225 + #228

## Description

When the query error is a 403, the component calls `setTimeout(() => navigate('/dashboard'), 3000)` inside the render path (not inside a useEffect). In React 18 StrictMode this is called twice per mount. The second `setTimeout` fires 3 seconds after mount with a stale closure but this is benign since `navigate` is stable. However, calling `setTimeout` in a render function is a side-effect that should live in `useEffect`.

## Fix

Move the 403 navigation side-effect into a `useEffect`:

```ts
const is403 = queryError instanceof TranslationServiceError && queryError.statusCode === 403;
useEffect(() => {
  if (!is403) return;
  const timer = setTimeout(() => navigate('/dashboard'), 3000);
  return () => clearTimeout(timer);
}, [is403, navigate]);
```

This also ensures the timer is cleared if the component unmounts before the 3 seconds elapse.

## Impact

Low — the current behaviour is functionally correct in production (StrictMode disabled). The timer is never cleared on unmount so if the user navigates away manually before 3 seconds, the `navigate('/dashboard')` fires anyway. Tracked as cleanup.

Issue #235, opened 2026-05-10
AuthContext StrictMode double-mount: loadUser() races in development


leixiaoyu:
## Finding

**File:** `frontend/src/contexts/AuthContext.tsx`
**Severity:** Medium
**Detected in:** OMC R1 for #225 + #228

## Description

React 18 StrictMode mounts effects twice in development. `AuthProvider` has a `useEffect(loadUser, [])` that fires twice concurrently. Both calls:

1. Read `getAuthToken()` — same token
2. Call `authService.getCurrentUser()` in parallel
3. Race to call `setUser(...)` and `setIsLoading(false)`

In production (no StrictMode), this is a non-issue. In development, if the first call fails (e.g. 401) and the second succeeds, the final state is correctly authenticated — but we burned an extra `/auth/me` request and potentially flashed an error state.

## Fix

Add an `isMounted` / `AbortController` guard inside `loadUser()` to cancel the second mount's in-flight request when the effect is cleaned up:

```ts
useEffect(() => {
  let cancelled = false;
  async function loadUser() {
    if (!getAuthToken()) return;
    try {
      const user = await authService.getCurrentUser();
      if (!cancelled) setUser(user);
    } catch (err) {
      if (!cancelled) { await authService.logout(); setUser(null); }
    } finally {
      if (!cancelled) setIsLoading(false);
    }
  }
  loadUser();
  return () => { cancelled = true; };
}, []);
```

## Impact

Low in production (StrictMode disabled). Medium in development — extra network requests and potential transient error flashes.

PR #234, opened 2026-05-10
fix(api): strip trailing slash from BASE_URL to prevent double-slash URLs (#224)

leixiaoyu:
## Summary

Resolves #224. During the 2026-05-09 E2E sweep we observed a request:

\`\`\`
POST https://8brwlwf68h.execute-api.us-east-1.amazonaws.com/v1//auth/refresh
\`\`\`

API Gateway tolerates the double slash today (returns 200), but the URL is malformed. Stricter gateways or future custom domains with path-rewrite rules can 404 on it. Other refresh requests in the same session use the correct `/v1/auth/refresh` form, indicating inconsistent URL composition rather than a missing endpoint.

## Root cause

`API_CONFIG.BASE_URL` in `frontend/src/config/constants.ts` was `getRequiredEnv('VITE_API_URL')` verbatim. When `VITE_API_URL` carries a trailing slash (the project's own `CLAUDE.md` documented it that way), the 401-refresh interceptor at `frontend/src/utils/api.ts:592` constructs:

\`\`\`ts
\`\${API_CONFIG.BASE_URL}/auth/refresh\`
\`\`\`

…yielding `/v1//auth/refresh`. The `authService.ts` path (`apiClient.post('/auth/refresh', ...)`) was at risk too because axios joins `baseURL + path` naively when both edges have slashes.

## Fix

One-line change in `constants.ts`:

\`\`\`ts
BASE_URL: getRequiredEnv('VITE_API_URL').replace(/\\/+$/, '')
\`\`\`

Single source of truth, zero per-callsite changes, defensive against any number of trailing slashes (`/v1`, `/v1/`, `/v1////` all normalize to `/v1`).

## Tests

Five regression-guard tests added in `api.test.ts`:

- BASE_URL with trailing slash is normalized
- BASE_URL with multiple trailing slashes is normalized
- BASE_URL without trailing slash is unchanged
- `apiClient.get()` produces no `//` in the constructed URL when env has trailing slash
- `apiClient.post()` likewise

OMC R1 self-review caught that the adapter-based tests were initially **vacuous** (they used `.env.test`'s slash-free `VITE_API_URL` and would have passed even with the fix reverted). Hardened with `vi.stubEnv` + `vi.resetModules()` inside each test to force the trailing-slash scenario; `vi.unstubAllEnvs()` + `vi.resetModules()` in `afterEach` to prevent stub bleed. The fixed tests now actually exercise the bug scenario.

## Conflict-with-merged-PR-#218 check

[...]

PR #238, opened 2026-05-10
fix(ui): auto-poll TranslationDetail on mount + auth hydration race (#225, #228)

### LINKED ISSUE #225: P1: Translation Details page doesn't auto-poll on first load — investor sees blank state
## Summary

After submitting a translation on /translation/upload, the SPA navigates to /translation/{jobId}. On arrival, the **Translation Progress card is not rendered** — only the static \"Job Information\" block (job ID, file name, target language, etc.) and a \"Start Translation\" button. No status, no progress bar, no chunk count.

The translation is **actually running** in the background (POST /translate succeeds, polling fires), but the user can't tell. They have to click \"Refresh Status\" to see anything.

Once Refresh Status is clicked, the full Translation Progress card materializes and updates correctly (verified end-to-end with pride-ch1.txt → Spanish, 22s, COMPLETED, 100%, 1/1 chunks, $0.00011).

## Severity

**P1** — demo polish, investor-visible. \"Click here to see if it's working\" is the wrong UX for a serverless real-time translation pitch.

## Repro

1. Local frontend on PR #218 + deployed dev backend (or after #218 deploys, deployed dev)
2. Login → Upload → wizard → Submit & Start Translation
3. Land on /translation/{id} — observe: no progress card, no status, just \"Start Translation\" button
4. Click \"Refresh Status\" → progress card appears, shows correct state

## Root cause hypothesis

\`frontend/src/pages/TranslationDetail.tsx\` does not invoke the polling loop on initial render. PR #125 added React Query adaptive polling per the Phase 10 status doc, but the trigger appears to require an explicit refresh action. Likely a `useEffect` that should kick a fetch on mount + poll until terminal state.

## Suggested fix

- On mount, immediately call `getJobStatus(jobId)` and start the polling loop (reuse the same adaptive cadence introduced in PR #125).
- Render the Translation Progress card unconditionally — show \"Loading…\" skeleton while the first fetch is in flight; render the real status once data lands.
- Hide \"Start Translation\" if \`translationStatus\` is anything other than the explicit pre-start state (NOT_STARTED) — currently it appears even when translation is already IN_PROGRESS / COMPLETED.

## Boy-scout follow-ups

- The \"Job Information\" block has an empty Content Type field (the API does return contentType on the response we audited — \`text/plain\` was missing on this job because the upload set fileSize but not contentType). Tracked separately if not addressed in the same PR.
- \`chunksTranslated\` is returned by the backend as a STRING (\`\"1\"\`) not a number — the page renders \"1 / 1\" which happens to coerce, but other downstream consumers may break. Tracked separately as backend hygiene.

### LINKED ISSUE #228: P2: ProtectedRoute bounces to /login on direct-URL navigation immediately after successful login (auth hydration race)
## Summary

Reproducible flow on http://localhost:3000 (frontend on PR #218 branch + deployed dev backend):

1. Visit /login
2. Enter valid credentials, click Log In
3. App routes to /dashboard, renders \"Welcome back, <name>!\" — login succeeded, session in localStorage (4213 bytes)
4. Programmatically (or by typing in the address bar) navigate to /translation/upload
5. **Bounced to /login** despite the session being present and valid

This does NOT reproduce when navigating via in-app SPA clicks (e.g. clicking \"Upload Document\" on the dashboard) — only on a full page load.

## Severity

**P2** — affects deep-linking, browser refresh, and any flow that re-enters the SPA on a protected route. Investor demo will run through happy-path SPA clicks and won't hit it, but a sharp observer who refreshes the page mid-flow will see a session loss they shouldn't.

## Root cause hypothesis

\`ProtectedRoute\` likely checks an in-memory auth flag that's still \`false\` on the very first render after a hard reload, before \`AuthContext\` has read localStorage and emitted the user. Classic hydration race.

The fix patterns to consider (pick one):

- Render a \`<Loading />\` from \`AuthContext\` until the localStorage read + initial \`/auth/me\` probe resolves; \`ProtectedRoute\` only decides AFTER \`isInitialized: true\`.
- Or read localStorage **synchronously** in the AuthContext constructor / module init so the very first render sees the session.

## Repro (with PR #218 worktree)

\`\`\`
cd .claude/worktrees/test-218/frontend
VITE_API_URL='https://8brwlwf68h...amazonaws.com/v1' npm run dev -- --port 3000 --host localhost
# Browser:
# 1. visit http://localhost:3000/login, log in
# 2. observe /dashboard renders
# 3. paste http://localhost:3000/translation/upload into address bar
# 4. observe /login renders despite localStorage.lfmt_session being populated
\`\`\`

## Boy-scout follow-up

The token-refresh interceptor in \`frontend/src/utils/api.ts\` does the right thing on 401 but doesn't coordinate with the AuthContext hydration. Worth a holistic look at the auth-bootstrap sequence.

leixiaoyu:
## Summary

Bundles two related React lifecycle / data-fetch race conditions surfaced during the 2026-05-09 demo prep E2E sweep.

### #225 — TranslationDetail page didn't auto-poll on first load

**Root cause** (uncovered during OMC R1, was NOT just a missing useEffect):
\`TranslationDetail.tsx\` was running TWO parallel data sources:
1. \`useTranslationJob\` (React Query, PR #125 adaptive polling) — was firing fine
2. A local \`fetchJobDetails()\` state copy that gated the \`<TranslationProgress />\` render — was only repopulated on \"Refresh Status\" click

Result: React Query's polling worked correctly but its data was being IGNORED. Investor saw a blank Progress card on submit-then-redirect.

**Fix**: Removed the parallel \`fetchJobDetails()\` local state. \`TranslationDetail\` now uses \`useTranslationJob\` exclusively. Skeleton renders during the first fetch; \`<TranslationProgress />\` renders the live data once available. The \"Start Translation\" button is gated strictly on \`status === 'CHUNKED'\`.

### #228 — ProtectedRoute bounced to /login on hard-reload despite valid session

**Root cause**: \`AuthContext\` initialized \`isLoading: false\` and only set it \`true\` once the async \`loadUser()\` ran. \`ProtectedRoute\`'s very first render saw \`isLoading=false + user=null\` and immediately redirected.

**Fix** (Option A — safer than synchronous localStorage read because it lets the \`/auth/me\` probe complete before deciding): lazy \`useState\` initializer reads \`getAuthToken()\` synchronously so the first render correctly emits \`isLoading=true\` when a session token exists.

## Critical bugs caught by OMC R1

- **C1**: AuthContext init bug above (#228 root cause).
- **C2**: TranslationDetail had an unsafe \`as TranslationJob['targetLanguage'] & string\` cast that could pass \`undefined\` to the API. Replaced with explicit guard.

## Boy-scout work bundled

- Added missing \`@testing-library/dom\` to devDependencies.
- Removed unused \`TranslationJob\` import after type-cast elimination.
- Promoted \`formatDate\` / \`formatFileSize\` helpers to module scope (DRY + perf).
- Removed clunky \`void updatedJob\` pattern by deleting \`onComplete\` callback entirely.

## Follow-ups filed

- **#235** (P3): StrictMode double-mount in AuthContext fires \`loadUser()\` twice concurrently.
- **#236** (P3): \`setTimeout(navigate, 3000)\` lives inside the render path instead of a \`useEffect\`.

## OMC R1 self-review

Two-commit structure:
- \`045a926\` — implementation
- \`993e65d\` — OMC R1 corrections

Full OMC review posted as a separate PR comment for record-keeping.

## Test plan

- [x] type-check clean (no new errors)
- [x] lint clean
- [x] prettier clean
- [x] vitest: **36 test files / 758 passed / 14 skipped**
- [ ] Manual: submit a translation → land on detail page → Progress card visible IMMEDIATELY on first paint (no Refresh click needed)
- [ ] Manual: log in → paste \`/translation/{jobId}\` URL into address bar → stays on detail page (no /login bounce)

## Closes

- Closes #225
- Closes #228

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -28,6 +28,7 @@
       "devDependencies": {
         "@axe-core/playwright": "^4.11.0",
         "@playwright/test": "^1.56.1",
+        "@testing-library/dom": "^10.4.1",
         "@testing-library/jest-dom": "^6.4.2",
         "@testing-library/react": "^14.2.1",
         "@testing-library/user-event": "^14.5.2",
@@ -2160,7 +2161,6 @@
       "integrity": "sha512-o4PXJQidqJl82ckFaXUeoAW+XysPLauYI43Abki5hABd853iMhitooc6znOnczgbTYmEP6U6/y1ZyKAIsvMKGg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.10.4",
         "@babel/runtime": "^7.12.5",
@@ -2350,6 +2350,7 @@
       "version": "18.3.26",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-18.3.26.tgz",
       "integrity": "sha512-RFA/bURkcKzx/X9oumPG9Vp3D3JUgus/d0b67KB0t5S/raciymilkOa66olh78MUI92QLbEJevO7rvqU/kjwKA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@types/prop-types": "*",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -43,6 +43,7 @@
   "devDependencies": {
     "@axe-core/playwright": "^4.11.0",
     "@playwright/test": "^1.56.1",
+    "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.4.2",
     "@testing-library/react": "^14.2.1",
     "@testing-library/user-event": "^14.5.2",
--- a/frontend/src/components/Auth/__tests__/ProtectedRoute.test.tsx
+++ b/frontend/src/components/Auth/__tests__/ProtectedRoute.test.tsx
@@ -307,4 +307,39 @@ describe('ProtectedRoute - Security Tests', () => {
       expect(screen.queryByText('Protected Dashboard Content')).not.toBeInTheDocument();
     });
   });
+
+  // -------------------------------------------------------------------------
+  // Issue #228 regression — auth hydration race on hard-reload
+  // -------------------------------------------------------------------------
+
+  describe('Issue #228 — Auth hydration race on hard-reload', () => {
+    it('shows a loading spinner (not /login redirect) when isLoading=true and user=null', () => {
+      // This is the exact state emitted by AuthProvider on the FIRST render
+      // when a token IS present in localStorage: isLoading=true, user=null.
+      // Before the fix AuthProvider initialized isLoading=false, causing
+      // ProtectedRoute to redirect immediately.
+      renderProtectedRoute({ user: null, isLoading: true });
+
+      expect(screen.getByRole('progressbar')).toBeInTheDocument();
+      expect(screen.queryByTestId('login-page')).not.toBeInTheDocument();
+      expect(screen.queryByTestId('protected-content')).not.toBeInTheDocument();
+    });
+
+    it('renders protected content after hydration completes (isLoading flips false with user set)', () => {
+      // Simulate the second render: /auth/me resolved, user is populated.
+      renderProtectedRoute({ user: mockUser, isLoading: false });
+
+      expect(screen.getByTestId('protected-content')).toBeInTheDocument();
+      expect(screen.queryByTestId('login-page')).not.toBeInTheDocument();
+      expect(screen.queryByRole('progressbar')).not.toBeInTheDocument();
+    });
+
+    it('redirects to login after hydration completes with no user (expired token)', () => {
+      // /auth/me returned 401 → AuthProvider sets user=null, isLoading=false.
+      renderProtectedRoute({ user: null, isLoading: false });
+
+      expect(screen.getByTestId('login-page')).toBeInTheDocument();
+      expect(screen.queryByTestId('protected-content')).not.toBeInTheDocument();
+    });
+  });
 });
--- a/frontend/src/contexts/AuthContext.tsx
+++ b/frontend/src/contexts/AuthContext.tsx
@@ -96,7 +96,17 @@ interface AuthProviderProps {
  */
 export function AuthProvider({ children }: AuthProviderProps) {
   const [user, setUser] = useState<User | null>(null);
-  const [isLoading, setIsLoading] = useState<boolean>(false);
+  // Issue #228: initialise isLoading to `true` when a session token already
+  // exists in localStorage, so ProtectedRoute sees a loading state on the
+  // very first render (before the async /auth/me probe completes) rather
+  // than `isLoading=false + user=null`, which caused an immediate redirect
+  // to /login on hard-reload of any protected route.
+  //
+  // Option A chosen over Option B (synchronous localStorage read) because
+  // the session might be stale/expired — we still want the /auth/me probe
+  // to validate it before declaring the user authenticated. isLoading=true
+  // is the correct signal: "we have a token, we are verifying it, hold on."
+  const [isLoading, setIsLoading] = useState<boolean>(() => getAuthToken() !== null);
   const [error, setError] = useState<ApiError | null>(null);
 
   /**
@@ -107,13 +117,15 @@ export function AuthProvider({ children }: AuthProviderProps) {
     async function loadUser() {
       const token = getAuthToken();
 
-      // If no token, user is not authenticated
+      // If no token, user is not authenticated; isLoading was initialised
+      // false (no token path), so no state update needed.
       if (!token) {
         return;
       }
 
-      // If token exists, try to load user profile
-      setIsLoading(true);
+      // Token exists — isLoading was already set to true by the useState
+      // initialiser above, so we skip the redundant setIsLoading(true) call.
+      // We go straight to the /auth/me probe.
 
       try {
         const currentUser = await authService.getCurrentUser();
--- a/frontend/src/pages/TranslationDetail.tsx
+++ b/frontend/src/pages/TranslationDetail.tsx
@@ -34,15 +34,27 @@ import DownloadIcon from '@mui/icons-material/Download';
 import ArrowBackIcon from '@mui/icons-material/ArrowBack';
 import RefreshIcon from '@mui/icons-material/Refresh';
 import CompareArrowsIcon from '@mui/icons-material/CompareArrows';
-import {
-  translationService,
-  TranslationServiceError,
-  TranslationJob,
-} from '../services/translationService';
+import { translationService, TranslationServiceError } from '../services/translationService';
 import { TranslationProgress } from '../components/Translation/TranslationProgress';
 import { useTranslationJob } from '../hooks/useTranslationJob';
 import { getLanguageLabel, getToneLabel } from '../utils/translationLabels';
 import { FEATURE_FLAGS } from '../config/constants';
+
+// ---------------------------------------------------------------------------
+// Pure helpers — module-level so they are not recreated on every render.
+// ---------------------------------------------------------------------------
+
+function formatDate(dateString: string): string {
+  return new Date(dateString).toLocaleString();
+}
+
+function formatFileSize(bytes: number): string {
+  if (bytes === 0) return '0 Bytes';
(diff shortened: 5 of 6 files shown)


leixiaoyu:
## OMC R1 Self-Review (pre-team-review pass)

Conducted in worktree \`agent-afc16938\` against \`origin/main\` (post-#234 merge).

### Summary verdict
**READY** — Two Critical bugs caught and fixed in the agent's own implementation. All required-checks green locally.

### Category Pass/Fail Matrix

| Category | Result | Notes |
|---|---|---|
| Architecture / SOLID / KISS / DRY | PASS | Single source of truth (React Query \`useTranslationJob\` hook); no duplicate polling primitive |
| Type Safety | PASS (after R1) | Removed unsafe \`& string\` cast; added runtime guards for optional fields |
| Test Coverage | PASS | 36 test files / 758 tests; regression tests assert behaviour not implementation |
| Race Conditions | PASS (after R1) | #228 hydration race fixed; StrictMode double-mount filed as #235 (medium) |
| Boy-Scout | PASS | Module-level helpers, removed \`onComplete\` void clunk, \`@testing-library/dom\` dep |
| PR Alignment | PASS | No conflict with PR #232 RegisterPage/LoginPage changes |

### Findings Table

| ID | Severity | File | Finding | Action |
|---|---|---|---|---|
| C1 | **Critical** | \`AuthContext.tsx\` | \`isLoading\` initialized \`false\` before async \`loadUser()\` could set it \`true\` — \`ProtectedRoute\` redirected immediately on hard-reload. | **Fixed**: lazy \`useState(() => getAuthToken() !== null)\` |
| C2 | **Critical** | \`TranslationDetail.tsx\` | \`as TranslationJob['targetLanguage'] & string\` unsafe cast; \`undefined\` could flow to API. | **Fixed**: explicit guard before \`startTranslation\` call |
| H1 | High | \`TranslationDetail.tsx\` | Page used parallel \`fetchJobDetails()\` local state alongside React Query — Progress card only rendered after manual Refresh click (this was the actual #225 root cause). | **Fixed**: page now exclusively uses \`useTranslationJob\`; skeleton on first paint |
| H2 | High | \`TranslationDetail.tsx\` | \"Start Translation\" button visibility | Verified correct: \`isChunked\` gate was already status-only; re-confirmed in tests |
| M1 | Medium | \`AuthContext.tsx\` | StrictMode double-mount fires \`loadUser()\` twice concurrently | Filed **#235** |
| M2 | Medium | \`TranslationDetail.tsx\` | \`setTimeout(navigate, 3000)\` inside render path instead of \`useEffect\` | Filed **#236** |
| M3 | Medium | \`TranslationDetail.tsx\` | \`void updatedJob\` pattern in \`onComplete\` callback | **Fixed**: removed \`onComplete\` entirely |
| M4 | Medium | \`TranslationDetail.tsx\` | \`formatDate\`/\`formatFileSize\` recreated per render | **Fixed**: promoted to module-level |

### Test Results

\`\`\`
type-check: no new errors
lint:       0 warnings, 0 errors
prettier:   clean
vitest:     36 test files / 758 passed / 14 skipped
\`\`\`

### Final commit SHAs

- \`045a926\` — implementation
- \`993e65d\` — OMC R1 corrections

### Boy-scout work included

- Added \`@testing-library/dom\` to \`devDependencies\` (was missing from worktree, caused 0 tests collected in TranslationDetail suite — would have hidden the H1 catch).
- Removed unused \`TranslationJob\` import from \`TranslationDetail.tsx\` after type-cast elimination.
- Promoted pure helpers to module scope (DRY + perf).

### Lesson worth capturing

The OMC R1 H1 finding was the actual #225 root cause — the original ticket described the symptom (\"Progress card not rendered until Refresh click\") as if the polling itself was broken. The deeper bug was \"polling worked fine; its data was being shadowed by a parallel local-state copy.\" Worth noting for future React Query / local-state coexistence reviews.

leixiaoyu:
## CI Fix — round 2

Pushed \`7051174\`. The previous push's CI failed only on a strict-mode TypeScript narrowing miss in OMC R1's C2 fix — the runtime guard narrowed \`string | undefined\` to \`string\`, but the API call expects the literal union \`'es' | 'fr' | 'de' | 'it' | 'zh'\`.

**Fix**: introduced module-level type predicates \`isSupportedLanguage\` and \`isSupportedTone\` so the runtime check produces the correct narrowed type at the call site. Avoids the project's prior \`as any\` pattern in \`TranslationUpload.tsx\`. Imports \`TranslationConfig\` as the SSoT — if a new language ships, the predicate type-errors until the supported-list is updated, providing a compile-time drift guard.

**No behavior change**: in-vocab values (es/fr/de/it/zh, formal/informal/neutral) accepted exactly as before; out-of-vocab values would have been rejected by the API anyway and now show a clear \"configuration is invalid\" message instead.

Local verification:
- type-check clean
- lint clean
- prettier clean
- vitest: 36 files / 758 passed / 14 skipped

xlei-raymond (approved):
As a Senior Reviewer, I have audited PR #238. This PR is a textbook example of "Codebase Stabilization" during a freeze—it doesn't just patch symptoms; it aggressively identifies and removes the architectural patterns (parallel state, incorrect initializations) that caused the
  bugs in the first place.

  Here is my detailed audit.

  ---

  1. Data Strategy: The Death of Parallel State
  The most impactful change is the refactoring of TranslationDetail.tsx.

  Educational Note:
  Previously, the component was trying to manage its own "local" copy of the job data alongside React Query. This is a common source of State Divergence. If React Query polls new data, but your local state doesn't update, the UI lies to the user.
   * The Lesson: Trust your "Single Source of Truth" (SSoT). If you are using a powerful cache manager like React Query, do not shadow its data in useState. The removal of fetchJobDetails() in favor of direct consumption of useTranslationJob() is the correct architectural move.

  2. Lifecycle Orchestration: Solving the "Login Bounce"
  The fix for #228 in AuthContext.tsx is subtle but critical for production-grade SPAs.

  Critique:
  Initializing isLoading: false when you know you have a token but haven't verified it yet is a race condition by design. ProtectedRoute (and other consumers) will interpret isLoading=false + user=null as "unauthenticated" and trigger redirects before the async verification
  completes.
   * The Lesson: Your initial state must reflect your potential state. If a token exists in localStorage, the system is "Loading (Verifying)" until proven otherwise. This prevents the jarring redirect-and-bounce that plagues many junior React apps.

  3. Type Safety: Eliminating Dangerous Casts
  I am particularly pleased with the OMC R1 fix regarding the targetLanguage and tone casts.

  Critique:
  The previous code used an as cast to force the API call. In TypeScript, as is often an admission of defeat. If the job object happened to be missing those fields (possible if the record was corrupted or still loading), we would have sent undefined to the backend, causing a 400
  or 500.
   * The Lesson: Use Type Predicates (isSupportedLanguage) and explicit guards instead of casts. This ensures that the code fails gracefully with a user-visible error rather than crashing the network layer.

  4. Pragmatic Refactoring: Hoisting Helpers
  Hoisting formatDate and formatFileSize to the module scope is a small but important "Boy Scout" win.

  Educational Note:
  Defining functions inside a component body means they are recreated on every single render. While the performance cost is negligible for two small functions, it creates unnecessary pressure on the garbage collector and can break memo or useEffect dependencies if those functions
  are ever passed down as props.
   * The Lesson: If a function doesn't depend on component state or props, it belongs outside the component.

  5. Testing the "Invisible"
  The update to TranslationDetail.test.tsx to assert on the progress-skeleton is excellent.

  Observation:
  We often forget to test our loading states. By verifying that the skeleton appears before the first fetch resolves, we ensure that the "First Meaningful Paint" for the investor demo will look polished and intentional, rather than a blank screen that "pops" into existence.

  ---

  Minor Technical Debt Note (The "P3" Follow-up):
  In TranslationDetail.tsx, the setTimeout inside the render path for 403 handling is technically a side-effect during render.
   * Senior Advice: In a post-demo refactor, this should be moved into a useEffect that watches for a forbidden state. However, given we are in a Code Freeze and the component is already returning an error UI at that point, I accept this as a "pragmatic compromise."

  Final Verdict:
  APPROVE & MERGE.

  The PR is architecturally superior to the code it replaces. It solves the reported demo blockers while simultaneously cleaning up the surrounding "code rot."

  Team: Study the AuthContext initialization pattern—it's the industry standard for handling persisted sessions in React.

  Merge when ready.

Issue #255, opened 2026-05-12
security: migrate auth tokens from localStorage to httpOnly cookies (Wave 2 Track D split from #197)


leixiaoyu:
## Context

Filed as the auth-architecture half of the original #197. The Wave 2 Track D PR (CSP-builder refactor #216 + violation telemetry #201 + contract guard #219) intentionally **defers** this concern with **strong data-points** per the fold-inline-or-justify-deferral policy:

### Strong data-point #1 — different concern class

This is **authentication architecture**, not CSP hardening. The Wave 2 Track D PR is scoped to CSP. The original #197 bundled these because both reduce XSS-exfiltration risk, but the implementation surfaces are completely disjoint:

| Aspect | CSP style-src nonces | httpOnly cookies |
| --- | --- | --- |
| Files touched | CDK CSP builder, Emotion CacheProvider | Lambda authorizers, axios client, CORS config, login/refresh/me handlers |
| AWS service surface | CloudFront response headers | API Gateway + Cognito + custom Lambda authorizer |
| Frontend churn | One CacheProvider wrap | Every axios call (`withCredentials`), every auth interceptor, login form |
| Test surface | CSP regression test | Every authenticated-endpoint integration test |

### Strong data-point #2 — cross-domain topology blocker

`SameSite=Lax` cookies do NOT cross from `d39xcun7144jgl.cloudfront.net` to `8brwlwf68h.execute-api.us-east-1.amazonaws.com` reliably across modern browsers (Safari ITP, Firefox ETP, Chrome's 3P-cookie deprecation). The clean fix is a **custom domain** for both (e.g., `app.lfmt.example.com` + `api.lfmt.example.com`) so the cookies are first-party. That's a separate operational change with ACM/Route53 implications — explicitly enumerated in the original #197 description.

### Strong data-point #3 — owner design-decision blocker

Several design decisions need an owner call before implementation can start:
- Custom-domain naming convention.
- CSRF mitigation: double-submit cookie vs `X-Requested-With` header vs both.
- Refresh-token cookie SameSite: `Strict` (safer, breaks third-party links) vs `Lax`.
- Cognito flow: does the Lambda authorizer extract the cookie and re-inject as Bearer, or does the integration receive the cookie directly?

Each of these is a 1–2 hour owner discussion. Bundling them into a code-PR drags the review chain.

## What's done in the Wave 2 Track D PR

[...]

Issue #254, opened 2026-05-12
security: CSP style-src nonces via build-time static nonce (Wave 2 Track D deferral from #197)


leixiaoyu:
## Status: Pivoted to build-time static-nonce architecture (2026-05-14)

> **Reader note**: the original Lambda@Edge `viewer-response` design was investigated and found **not implementable as specified** — Lambda@Edge response triggers do NOT expose the upstream response body (only `headers`/`status`/`statusDescription`). The full investigation, AWS doc citations, and pivot rationale are preserved in [comment #4456187177](https://github.com/leixiaoyu/lfmt-poc/issues/254#issuecomment-4456187177). The original design (kept for historical context) is at the bottom of this body under "Original (superseded) design".
>
> The acceptance criteria below reflect the **active scope**.

## Updated architecture — build-time static-nonce + CDK custom resource

A CDK custom resource generates a random nonce at `cdk deploy` time. The same nonce is:

1. Written into the S3-uploaded `index.html` (replacing a `__CSP_NONCE__` placeholder set by Vite).
2. Emitted via CloudFormation output/parameter and used in the response-headers policy's CSP string for `style-src 'nonce-<value>'`.

The nonce is static for the lifetime of one deploy — rotates on every frontend deploy.

### Trade-off vs. the original per-response design

The nonce is static within one deploy lifetime. An attacker who reads `index.html` learns the nonce until the next frontend deploy. This is a meaningful but bounded reduction in the security benefit.

### Why this is the right call now

- The CSP violation-report endpoint (#201, live since PR #257) catches any regression that introduces inline styles without nonces — i.e., this design's main goal (eliminate `'unsafe-inline'`) is achievable + alarmable.
- The remaining auth-token surface is still localStorage-readable (#255 deferred); per-response nonce freshness on style-src is comparatively low marginal value while that exposure exists.
- Strong-data-point #2 below already enumerated the deploy-sync constraint as a blocker. The static-nonce path addresses it directly via the CDK custom resource emitting the nonce the backend reads.

### Acceptance criteria (active)

- [ ] CDK custom resource generates a deploy-time random nonce.
- [ ] Vite build emits `__CSP_NONCE__` placeholder in `index.html` (no behavior change at build time).
- [ ] CDK custom resource reads built `index.html` from S3 upload bucket, replaces placeholder with the nonce, re-uploads (one-time at deploy).
[...]

PR #257, opened 2026-05-12
feat(security): Wave 2 Track D — CSP refactor + violation telemetry + live contract guard (#216 #201 #219)

### LINKED ISSUE #201: security: add CSP report-uri/report-to endpoint for violation telemetry
## Context

PR #198 (Issue #194) tightened the CSP — \`script-src\` is now \`'self'\` only (no \`'unsafe-inline'\`), and a regression test in \`infrastructure.test.ts\` blocks reintroduction. However, the CSP currently has NO reporting endpoint, so a future regression that violates the policy in production would silently break user requests without telemetry.

OMC review (security-auditor) flagged this as a Recommended item. Implemented via two CSP directives:
- \`report-uri <endpoint>\` — legacy directive, still supported by all browsers.
- \`report-to <group>\` — modern Reporting API; requires a \`Report-To\` HTTP header that defines the group.

### Option A: Lightweight Lambda + CloudWatch (recommended for POC)

1. New Lambda function \`csp-report-collector\` that:
   - Accepts \`POST /csp-report\` with the standard CSP violation report body.
   - Validates the \`content-type\` is \`application/csp-report\` or \`application/reports+json\`.
   - Logs to CloudWatch with structured fields (\`violated-directive\`, \`blocked-uri\`, \`document-uri\`, \`source-file\`, \`line-number\`, \`user-agent\`).
   - Rate-limits per source IP to prevent log flooding (e.g., 10/min/IP using DynamoDB token bucket — the existing distributed rate-limiter pattern works).
2. New API Gateway route \`POST /csp-report\` that maps to the Lambda. NO authentication (browsers send these reports anonymously).
3. CloudWatch metric filter on the log group: count violations, alarm on >0 per hour for \`script-src\` violations.
4. CDK CSP definition in \`lfmt-infrastructure-stack.ts\`: append \`report-uri https://<api-domain>/csp-report\` to BOTH CSP-carrying response-headers policies. The \`report-to\` directive can be added later once we agree on a Reporting API group name.

### Option B: Third-party CSP report collector

Free tiers exist (e.g., report-uri.com, csper.io). Trade-off: adds a third-party dependency, but zero infrastructure overhead. Probably overkill for a POC.

## Why this matters

- A legitimate user-blocking regression (e.g., a future PR re-adds an inline \`<script>\` in \`index.html\`) would otherwise only show as a CSP-violation entry in each user's browser console — invisible to ops.
- \`script-src 'self'\` is a fail-closed directive; a violation = a broken page. Telemetry is essential to catch regressions BEFORE users complain.
- The same endpoint will collect future violations for the deferred \`style-src\` nonce work (#197), so the endpoint pays for itself.

## References

- PR #198 — added the \`script-src\` regression test that this telemetry would back up
- Issue #194 — original CSP-tightening risk
- Issue #197 — deferred CSP/auth hardening (style-src nonces, httpOnly cookies)

### LINKED ISSUE #216: refactor(infra): evolve buildCsp options to directives?: Partial<Record<CspDirective, string[]>> shape
## Context

Filed as a follow-up from PR #214 OMC R2 — the Architect specialist (S2/R3) flagged that the current `buildCsp` options shape:

```ts
private buildCsp(opts: { connectSrc: string[]; reportUri?: string }): string
```

…will not compose cleanly with the upcoming **#197 style-src nonce** work. Once Lambda@Edge starts injecting per-response nonces into `style-src` (and possibly `script-src`), every new directive that needs runtime-derived input would have to be added as a new named field — `connectSrc`, `reportUri`, `styleSrcNonce`, `scriptSrcNonce`, … — and the call site (`updateCloudFrontCSP`) would grow a corresponding parameter for each.

## Proposed Refactor

Replace the per-directive named fields with a single, structured map:

```ts
type CspDirective =
  | 'default-src'
  | 'script-src'
  | 'style-src'
  | 'img-src'
  | 'font-src'
  | 'connect-src'
  | 'object-src'
  | 'base-uri'
  | 'form-action'
  | 'frame-ancestors'
  | 'report-uri';

private buildCsp(opts: {
  directives?: Partial<Record<CspDirective, string[]>>;
}): string
```

The defaults (current hardening list) move into `buildCsp`'s body; callers only pass the directives they want to override or extend.

## Why Now (Spec Pointer)

- PR #214 OMC R2 architect S2 + R3 — the current shape is "single-purpose" and anticipates only `report-uri` (#201). Adding nonces will require a fan-out of named parameters that quickly becomes hard to audit.
- Issue #197 (style-src nonce composition) is the first real consumer that will exercise the cross-directive composition contract.

## Acceptance Criteria

- [ ] `buildCsp` accepts a `directives?: Partial<Record<CspDirective, string[]>>` map.
- [ ] H-3 reportUri sanitization (added in PR #214 commit) preserved against the new shape.
- [ ] All existing call sites migrated; no behaviour change in the synthesized CSP.
- [ ] Unit tests cover: default emission, single-directive override, multi-directive override, and nonce-style runtime values.

## Out of Scope

- Implementing the actual style-src nonce pipeline — that lives in #197.
- Adding a CSP-violation receiver Lambda — that lives in #201.

Linked: #197.

### LINKED ISSUE #219: test(e2e): live-backend API envelope contract guard against deployed dev
## Context

PR #218 (`fix/hotfix-api-envelope-mismatch`) introduced an MSW-driven contract guard in `frontend/src/__tests__/apiEnvelopeContract.test.ts`. That guard validates the wire envelope against the **mock** handlers (which now mirror the real Lambda response shape via the shared `*ApiResponse` interfaces in `@lfmt/shared-types`).

OMC R1 review test C5 raised the gap: the mock could still drift from the deployed Lambda (e.g. a manual AWS-console change, a production hot-fix that bypasses CDK, a shared-types version mismatch in the deployed bundle). We need a Playwright spec that:

1. Authenticates against the deployed dev backend.
2. Calls each protected endpoint (POST /jobs/upload, POST /jobs/{id}/translate, GET /jobs/{id}/translation-status, GET /jobs, etc.).
3. Asserts the response shape matches the corresponding `@lfmt/shared-types` interface AND the helper used (`createFlatResponse` vs `createWrappedResponse`).
4. Runs in CI on a schedule (post-deploy nightly is fine — every PR is too noisy and risks flaky-network failures gating merges).

## Acceptance criteria

- New Playwright spec under `frontend/tests/e2e/contract/` (or equivalent) that asserts wire shapes against the deployed dev API.
- Reads endpoint URL from `VITE_API_URL` (or a dedicated `E2E_API_URL`).
- Documented invocation pattern in `frontend/LOCAL-TESTING.md` (or a sibling doc).
- Wired into a scheduled GitHub Actions workflow (e.g. `e2e-contract-nightly.yml`) — NOT the per-PR CI workflow.

## Reference

PR #218 OMC R1 test C5 (envelope contract guard against deployed dev).

leixiaoyu:
## Summary

Wave 2 Track D of the tech-debt cleanup. **3 issues resolved + 1 split-with-strong-justification** (#197 split into #254 + #255). Touches CSP construct, new violation-report Lambda, frontend e2e contract guard, scheduled GitHub Actions workflow.

## Issues resolved

- **#216** — Evolved `buildCsp` options to typed `Partial<Record<CspDirective, string[]>>` shape. Extracted CSP builder into standalone `backend/infrastructure/lib/csp.ts`. Preserves the previous H-3 reportUri sanitization + adds a CDK-token escape hatch with defense-in-depth tests (protocol + char checks run BEFORE token detection so a `;`-injection or `http://`-prefix token URL is rejected).
- **#201** — New `POST /csp-report` Lambda (`backend/functions/security/cspReport.ts`) with strict input sanitization:
  - Content-Type allowlist (`application/csp-report` + `application/json`)
  - 64 KB body cap (pre-parse DoS guard)
  - Field allowlist on log emission (no log poisoning via attacker-controlled keys)
  - Per-field 2 KB truncation
  - 204-on-success, no body echo
  - Dedicated IAM role: basic-execution ONLY (no DDB, no SSM, no anything else)
  - 25 unit tests covering all validation paths
- **#219** — Live-backend API envelope contract guard. New Playwright spec at `frontend/e2e/tests/contract/api-envelope-live.spec.ts` authenticates against deployed dev, calls every protected endpoint, asserts response shape matches the corresponding `@lfmt/shared-types` `*ApiResponse` interface. New scheduled GitHub Actions workflow at `.github/workflows/e2e-contract-nightly.yml` — **NOT** per-PR gate per issue body. Credentials from CI secrets (`LFMT_TEST_EMAIL` / `LFMT_TEST_PASSWORD`), per-run-user fallback via Cognito auto-confirm.

## Split with strong-justification

**#197** (style-src nonces + httpOnly cookies) is two concerns:

| Original | Where it landed | Strong data-points |
|---|---|---|
| Style-src nonces | **#254** (new follow-up) | Wrong architectural layer — per-response nonces require Lambda@Edge, explicitly out-of-scope per issue body. Different concern class (CloudFront operational architecture, not CSP-builder refactor). Frontend/backend deploy sync constraint documented in `docs/CI-CD-ARCHITECTURE.md`. |
| httpOnly cookies for auth tokens | **#255** (new follow-up) | Different concern class — auth architecture, not CSP. Cross-domain topology blocker: SameSite=Lax cookies don't reliably cross `cloudfront.net` → `execute-api.amazonaws.com`; needs custom-domain ACM/Route53. Owner design-decision blockers — CSRF mitigation choice, SameSite mode, Cognito authorizer integration shape. |

Both new issues have full justifications cited per user policy.

## OMC R1 self-review

Conducted in a separate commit. No Critical/High findings. Medium findings (CDK token escape hatch, IAM minimality, log-poisoning anti-forgery) all addressed inline.

## Boy-scout work included

- `assertValidCspReportUri` now rejects non-string input with a clear message
- New `report-uri rejects empty array` test prevents future sourceless-directive grammar error
- `EXPECTED_APPLICATION_LAMBDA_COUNT` drift-guard bumped 15 → 16 with comment trail (new cspReport Lambda)

## Test plan

- [x] `cd backend/infrastructure && npm run build && npm test` — **89 passed**
- [x] `cd backend/functions && npm test` — **589 passed, 3 skipped**
- [x] `cd frontend && npm run type-check && npm run lint && npx prettier --check .` — all clean
- [x] `cd frontend && VITE_API_URL=... CI=true npm run build` — clean
- [x] `cd backend/infrastructure && npx cdk synth --context environment=dev` — CSP emits `report-uri https://<api>/v1/csp-report;` as LAST directive
- [ ] Manually trigger `.github/workflows/e2e-contract-nightly.yml` against deployed dev (post-merge)
- [ ] Verify malformed `POST /csp-report` returns 400 with NO CloudWatch entry (anti log-flood)
- [ ] Verify valid report returns 204 with ONE structured WARN entry

## Coordination note

Touches `backend/infrastructure/lib/lfmt-infrastructure-stack.ts` for the new Lambda + route. Wave 2 Track B PR (in flight in parallel) also touches that file for #178/#210 IAM changes. If this PR lands first, Track B will need to rebase on `EXPECTED_APPLICATION_LAMBDA_COUNT` (15 → 16 bumped here). Conflict expected to be small.

## Closes

- Closes #216
- Closes #201
- Closes #219

(#197 closed via split: #254 + #255 carry the two halves with strong-justification.)

### DIFF
--- a/.github/workflows/cognito-test-user-cleanup.yml
+++ b/.github/workflows/cognito-test-user-cleanup.yml
@@ -1,0 +1,217 @@
+name: Cognito Test User Cleanup (Weekly GC)
+
+# Weekly garbage-collection sweep for ephemeral test users created by the
+# live-backend contract spec (#219) when shared CI credentials
+# (LFMT_TEST_EMAIL / LFMT_TEST_PASSWORD) are absent.
+#
+# The fallback path in `frontend/e2e/tests/contract/api-envelope-live.spec.ts`
+# registers a fresh user per run with the pattern
+#   contract-<ts>-<rand>@e2e-test.lfmt.com
+# and the dev Cognito pool's PreSignUp auto-confirm trigger makes that
+# synchronous so the spec can log in immediately. In normal operation the
+# shared-credentials path is taken and this fallback fires only on forked
+# `workflow_dispatch` runs that lack secret access — but over a year that
+# accumulation still adds up.
+#
+# This workflow runs weekly and:
+#   1. Reads the dev User Pool ID from the LfmtPocDev CloudFormation stack
+#      output (single source of truth — no hardcoded pool IDs).
+#   2. Lists users whose email matches `contract-*@e2e-test.lfmt.com`.
+#   3. Deletes every match older than 7 days (the operational SLA for
+#      reproducing a nightly failure — we never want to GC a user that
+#      on-call might still need for triage).
+#
+# OMC R2 Medium-2 follow-up to PR #257 (Wave 2 Track D).
+#
+# ---------------------------------------------------------------------------
+# IAM requirements
+# ---------------------------------------------------------------------------
+#
+# The OIDC role assumed via `secrets.AWS_ROLE_ARN` must have:
+#   - cognito-idp:ListUsers          (scoped to dev User Pool ARN)
+#   - cognito-idp:AdminDeleteUser    (scoped to dev User Pool ARN)
+#   - cloudformation:DescribeStacks  (already present — deploy workflows
+#                                     use this for stack-output reads)
+#
+# The current deploy role grants the Cognito admin actions (it creates the
+# pool), so no IAM update is expected to be required. If a future tightening
+# narrows the deploy role, the cleanup pool ARN MUST stay on the allow-list
+# or this workflow will start failing every Sunday.
+
+on:
+  schedule:
+    # Sunday 10:00 UTC = 03:00 PT (winter) / 02:00 PT (DST). Avoid the
+    # 00:00–06:00 UTC window where GitHub's scheduler clusters runs.
+    - cron: '0 10 * * 0'
+  # Manual trigger for on-call to sweep on demand after a noisy fallback
+  # week.
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Dry run (list only, do not delete)'
+        required: false
+        type: boolean
+        default: false
+      max_age_days:
+        description: 'Delete users older than N days (default 7)'
+        required: false
+        type: string
+        default: '7'
+
+env:
+  AWS_REGION: us-east-1
+  # The dev stack name — must match `backend/infrastructure/bin/lfmt-cdk.ts`.
+  # If a future PR renames the stack, this value must change in lockstep.
+  STACK_NAME: LfmtPocDev
+  # Email pattern hardcoded to match the spec's `generateRunUser` template
+  # (api-envelope-live.spec.ts line ~106). Drift between this pattern and
+  # the spec would silently miss zombie users — keep them in sync.
+  EMAIL_PREFIX: 'contract-'
+  EMAIL_DOMAIN: '@e2e-test.lfmt.com'
+
+permissions:
+  # OIDC token issuance — required by aws-actions/configure-aws-credentials.
+  id-token: write
+  # Read-only repo access; this workflow does not push commits or comments.
+  contents: read
+
+# Serialize against a manual dispatch on the same day so we don't double-
+# delete a user mid-list.
+concurrency:
+  group: cognito-test-user-cleanup
+  cancel-in-progress: false
+
+jobs:
+  cleanup:
+    name: Cognito test user GC (dev pool)
+    runs-on: ubuntu-latest
+    # 10-min cap. Listing + deleting ~365 users (one year of nightly
+    # fallbacks) takes <2 min via the AWS CLI; anything longer means the
+    # pool is much larger than expected or the API is throttling — fail
+    # loudly rather than burn CI time.
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Configure AWS credentials (OIDC)
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          role-session-name: GitHubActions-LFMT-CognitoGC
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Verify AWS identity
+        run: aws sts get-caller-identity
+
+      - name: Resolve dev User Pool ID from stack output
+        id: pool
+        run: |
+          USER_POOL_ID=$(aws cloudformation describe-stacks \
+            --stack-name "${STACK_NAME}" \
+            --query "Stacks[0].Outputs[?OutputKey=='UserPoolId'].OutputValue" \
+            --output text)
+          if [ -z "${USER_POOL_ID}" ] || [ "${USER_POOL_ID}" = "None" ]; then
+            echo "::error::Could not resolve UserPoolId from stack ${STACK_NAME}"
+            exit 1
+          fi
+          echo "Resolved UserPoolId: ${USER_POOL_ID}"
+          echo "user_pool_id=${USER_POOL_ID}" >> "$GITHUB_OUTPUT"
+
+      - name: List + delete stale test users
+        env:
+          USER_POOL_ID: ${{ steps.pool.outputs.user_pool_id }}
+          DRY_RUN: ${{ inputs.dry_run || 'false' }}
+          MAX_AGE_DAYS: ${{ inputs.max_age_days || '7' }}
+        run: |
+          set -euo pipefail
+
+          # Cutoff timestamp in seconds (epoch). Users created on/after
+          # this timestamp are PRESERVED — on-call may still need them
+          # for nightly-failure triage.
+          NOW=$(date -u +%s)
+          CUTOFF=$(( NOW - (MAX_AGE_DAYS * 86400) ))
+          echo "Cutoff: ${MAX_AGE_DAYS} days = epoch ${CUTOFF}"
+          echo "Dry run: ${DRY_RUN}"
+
+          # AWS Cognito ListUsers supports a `Filter` parameter with
+          # limited syntax — `email ^=` is the documented prefix match.
+          # We additionally post-filter by domain suffix in shell so a
+          # future test pattern change doesn't accidentally GC users
+          # outside our test domain.
+          USERS_JSON=$(aws cognito-idp list-users \
+            --user-pool-id "${USER_POOL_ID}" \
+            --filter "email ^= \"${EMAIL_PREFIX}\"" \
+            --limit 60 \
+            --output json)
+
+          # Iterate paginated results. The AWS CLI exposes the next-page
+          # token in the response; loop until exhausted. Per-user
+          # DELETE/PRESERVE/SKIP records are emitted from `process_page`
+          # so the job log itself is the audit trail (no aggregate
+          # counters maintained — running them through a subshell loop
+          # would require fd-based bookkeeping for portability).
+          NEXT_TOKEN=$(echo "${USERS_JSON}" | jq -r '.PaginationToken // empty')
+
+          process_page() {
+            local page="$1"
+            echo "${page}" | jq -c '.Users[]' | while read -r user; do
+              local username email created
+              username=$(echo "${user}" | jq -r '.Username')
+              email=$(echo "${user}" | jq -r '.Attributes[] | select(.Name=="email") | .Value')
+              # UserCreateDate is ISO 8601 — convert to epoch.
+              created=$(echo "${user}" | jq -r '.UserCreateDate')
+
+              # Defensive double-check on the email domain. If a future
+              # test changes the pattern but we forget to update the
+              # workflow filter, the prefix `contract-` could match
+              # legitimate test accounts in production. The domain check
+              # narrows the blast radius.
+              if [[ "${email}" != *"${EMAIL_DOMAIN}" ]]; then
+                echo "SKIP (wrong domain): ${email}"
+                continue
+              fi
+
+              created_epoch=$(date -u -d "${created}" +%s 2>/dev/null \
+                || gdate -u -d "${created}" +%s 2>/dev/null \
+                || python3 -c "import sys, datetime as dt; print(int(dt.datetime.fromisoformat(sys.argv[1].replace('Z','+00:00')).timestamp()))" "${created}")
+
+              if [ "${created_epoch}" -ge "${CUTOFF}" ]; then
+                echo "PRESERVE (too young): ${email} (created ${created})"
+                continue
+              fi
+
+              if [ "${DRY_RUN}" = "true" ]; then
+                echo "DRY RUN — would delete: ${email} (created ${created})"
+              else
+                echo "DELETE: ${email} (created ${created})"
+                aws cognito-idp admin-delete-user \
+                  --user-pool-id "${USER_POOL_ID}" \
+                  --username "${username}"
+              fi
+            done
+          }
+
+          process_page "${USERS_JSON}"
+
+          while [ -n "${NEXT_TOKEN}" ]; do
+            USERS_JSON=$(aws cognito-idp list-users \
+              --user-pool-id "${USER_POOL_ID}" \
+              --filter "email ^= \"${EMAIL_PREFIX}\"" \
+              --limit 60 \
+              --pagination-token "${NEXT_TOKEN}" \
+              --output json)
+            process_page "${USERS_JSON}"
+            NEXT_TOKEN=$(echo "${USERS_JSON}" | jq -r '.PaginationToken // empty')
+          done
+
+          echo "Cleanup pass complete."
+
+      - name: Report summary
+        if: always()
+        run: |
+          echo "Cognito test user cleanup workflow finished."
+          echo "Stack:    ${STACK_NAME}"
+          echo "Pattern:  ${EMAIL_PREFIX}*${EMAIL_DOMAIN}"
+          echo "See job log above for per-user PRESERVE/DELETE/SKIP records."
--- a/.github/workflows/e2e-contract-nightly.yml
+++ b/.github/workflows/e2e-contract-nightly.yml
@@ -1,0 +1,119 @@
+name: E2E Contract Nightly (Live Backend)
+
+# Live-backend API envelope contract guard (#219).
+#
+# Runs the Playwright spec at
+#   `frontend/e2e/tests/contract/api-envelope-live.spec.ts`
+# against the deployed dev backend on a nightly schedule. Each test
+# authenticates with Cognito, hits a protected endpoint, and asserts the
+# response shape matches the corresponding `*ApiResponse` interface in
+# `@lfmt/shared-types`.
+#
+# Why NOT per-PR (issue #219 explicit requirement):
+# - Per-PR runs would burn AWS quota on every PR (Cognito user creation,
+#   S3 upload, Step Functions execution).
+# - The live network can flake; gating unrelated PRs on flaky-network
+#   failures is unhelpful.
+# - PRs that deploy backend changes during their own CI run would race
+#   with this spec.
+#
+# Triggers:
+# - `schedule` — 09:00 UTC daily (post-deploy window, before US business
+#   hours so a failure is visible at start-of-day).
+# - `workflow_dispatch` — on-call can re-run after a hot-fix without
+#   waiting for the next scheduled tick.
+
+on:
+  schedule:
+    # 09:00 UTC = 02:00 PT (winter) / 01:00 PT (DST). Avoid running on the
+    # exact :00 minute because GitHub schedules cluster there and queue.
+    - cron: '17 9 * * *'
+  workflow_dispatch:
+    inputs:
+      api_url:
+        description: 'API base URL (defaults to deployed dev API)'
+        required: false
+        type: string
+        default: ''
+
+env:
+  NODE_VERSION: '22'
+
+# Single in-flight run — nightly + a hot-fix dispatch on the same morning
+# would otherwise race on the shared test account's Cognito session.
+concurrency:
+  group: e2e-contract-nightly
+  cancel-in-progress: false
+
+jobs:
+  contract-live:
+    name: API envelope contract (live dev backend)
+    runs-on: ubuntu-latest
+    # 15 min hard cap — the spec hits 6 endpoints and should complete in
+    # under 2 min on a healthy backend. A long-running job indicates the
+    # deployed Lambda is hung; cancel-and-alarm rather than burn CI time.
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: 'npm'
+          cache-dependency-path: |
+            frontend/package-lock.json
+            shared-types/package-lock.json
+
+      - name: Install shared-types
+        working-directory: shared-types
+        run: |
+          npm ci
+          npm run build
+
+      - name: Install frontend deps
+        working-directory: frontend
+        run: npm ci
+
+      - name: Install Playwright browsers
+        working-directory: frontend
+        # `--with-deps` installs the OS-level browser dependencies on
+        # ubuntu-latest. Without it Chromium fails to launch with a
+        # missing-shared-library error.
+        run: npx playwright install --with-deps chromium
+
+      - name: Run live-backend contract spec
+        working-directory: frontend
+        # The spec reads credentials from env vars. NEVER hardcode them
+        # in the spec file — see the file-level comment in
+        # api-envelope-live.spec.ts.
+        env:
+          # Repo secrets: set under Settings → Secrets and variables →
+          # Actions. The spec falls back to per-run user registration
+          # when these are absent (Cognito auto-confirm makes that
+          # synchronous on dev), so a forked workflow_dispatch still
+          # works without secret access.
+          LFMT_TEST_EMAIL: ${{ secrets.LFMT_TEST_EMAIL }}
+          LFMT_TEST_PASSWORD: ${{ secrets.LFMT_TEST_PASSWORD }}
+          LFMT_API_URL: ${{ inputs.api_url || 'https://8brwlwf68h.execute-api.us-east-1.amazonaws.com/v1/' }}
+          # The contract spec doesn't navigate the SPA — it makes raw
+          # API calls — so PLAYWRIGHT_BASE_URL is irrelevant. Set it
+          # to a no-op value so the dev-server bootstrap in
+          # playwright.config.ts is skipped.
+          PLAYWRIGHT_BASE_URL: ${{ inputs.api_url || 'https://8brwlwf68h.execute-api.us-east-1.amazonaws.com/v1/' }}
+        # `--grep` filters to ONLY the contract-live tagged suite so a
+        # future addition to the e2e/ tree doesn't accidentally execute
+        # under the nightly slot.
+        run: npx playwright test --grep "@contract-live" --reporter=html,junit
+
+      - name: Upload Playwright report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report-contract-live
+          path: |
+            frontend/playwright-report
+            frontend/test-results
+          retention-days: 14
--- a/backend/functions/security/cspReport.test.ts
+++ b/backend/functions/security/cspReport.test.ts
@@ -50,9 +50,7 @@ describe('cspReport — normalizeContentType', () => {
     expect(normalizeContentType('Application/CSP-Report; charset=utf-8')).toBe(
       'application/csp-report'
     );
-    expect(normalizeContentType('application/reports+json')).toBe(
-      'application/reports+json'
-    );
+    expect(normalizeContentType('application/reports+json')).toBe('application/reports+json');
   });
 
   it('returns empty string on non-string input', () => {
--- a/backend/functions/security/cspReport.test.ts
+++ b/backend/functions/security/cspReport.test.ts
@@ -1,0 +1,402 @@
+/**
+ * Unit tests for the CSP Violation Report collector (#201).
+ *
+ * Covers:
+ *   1. Method/content-type/size gates (return 400 without parsing).
+ *   2. Legacy `application/csp-report` parse path — required fields,
+ *      field truncation, integer coercion, disposition allowlist.
+ *   3. Reporting API `application/reports+json` parse path — batched
+ *      reports, non-CSP entries dropped, camelCase field names.
+ *   4. Anti-log-poisoning — extra fields in payload are NOT forwarded.
+ *   5. CORS preflight returns 204 without invoking the parser.
+ */
+
+import type { APIGatewayProxyEvent } from 'aws-lambda';
+import {
+  handler,
+  parseLegacyReport,
+  parseReportsApiBody,
+  normalizeContentType,
+  type SanitizedCspReport,
+} from './cspReport';
+
+/** Build a minimal API Gateway event for POST /csp-report. */
+function buildEvent(opts: {
+  method?: string;
+  contentType?: string | undefined;
+  body?: string | null;
+}): APIGatewayProxyEvent {
+  return {
+    httpMethod: opts.method ?? 'POST',
+    headers: opts.contentType !== undefined ? { 'Content-Type': opts.contentType } : {},
+    body: opts.body ?? null,
+    path: '/csp-report',
+    isBase64Encoded: false,
+    multiValueHeaders: {},
+    pathParameters: null,
+    queryStringParameters: null,
+    multiValueQueryStringParameters: null,
+    stageVariables: null,
+    requestContext: {
+      requestId: 'test-req',
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any,
+    resource: '/csp-report',
+  };
+}
+
+describe('cspReport — normalizeContentType', () => {
+  it('lowercases and strips parameters', () => {
+    expect(normalizeContentType('Application/CSP-Report; charset=utf-8')).toBe(
+      'application/csp-report'
+    );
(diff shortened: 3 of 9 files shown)


leixiaoyu:
## OMC R1 Self-Review — Wave 2 Track D (CSP + violation telemetry + live contract guard)

**Reviewed commit**: `e3dfe82` (HEAD on `feat/wave2-track-d-security-csp`, includes the origin/main merge that brought in #250/#251/#252/#256).
**Diff base**: `f16303b` (branch-cut from `main` pre-Wave-1).
**Issues covered**: #216 (CSP builder refactor), #201 (violation-report Lambda), #219 (live contract guard). #197 split into #254 + #255 (filed as follow-ups; not closed by this PR).
**New tests vs base**: ~25 cspReport unit + ~10 csp.ts builder + new infrastructure assertions + new Playwright `@contract-live` suite + nightly workflow.
**Reviewer note**: this PR is INTERNET-FACING (the `POST /csp-report` route is `AuthorizationType.NONE` by design). The review is calibrated for that threat model.
**Date**: 2026-05-13

---

### Category pass/fail matrix

| Category | Result | Notes |
|---|---|---|
| Architecture / SOLID | **PASS** | `csp.ts` extracted as a pure module (no AWS/CDK deps); `cspReport.ts` has a clean 3-stage pipeline (validate → parse → emit) with each stage exported for isolated testing. Single source of truth for hardening defaults. |
| Security (anonymous endpoint) | **PASS** | 7-layer defensive model implemented and individually testable: content-type allowlist, body-size cap, field allowlist, length caps, no-log-on-400, 204-no-body-echo, fail-closed errors. |
| Security (CSP-string injection) | **PASS** | `assertValidCspReportUri` runs forbidden-char + protocol checks BEFORE `new URL()` so a `;`-injection or `http://`-prefixed CDK token URL is rejected even with the token escape hatch. Defense-in-depth tests pin both regressions. |
| Type safety | **PASS** | `CspDirective` union narrows the directive name set; `Required<Pick<...>>` on `DEFAULT_DIRECTIVES` forces a default for every required directive; `SanitizedCspReport` interface pins the log-record shape. No `any` in production code. |
| IAM | **PASS** | `CspReportLambdaRole` is a dedicated role with ONLY `AWSLambdaBasicExecutionRole`. No DDB/S3/SSM/SFN. The anonymous-internet exposure rationale is in the construct's `description`. |
| Test coverage | **PASS** | 89 infra tests / 598 functions tests / new `@contract-live` Playwright suite. CDK-token escape-hatch + every defense layer has at least one negative test. |
| CI/CD | **PASS** | New `e2e-contract-nightly.yml` uses `concurrency` to serialize against `workflow_dispatch`; 15-min timeout; explicit "why not per-PR" rationale in the workflow header. |
| Documentation | **PASS** | Every defensive choice has an inline `// Why:` comment; the file-level header on `cspReport.ts` is a complete security model summary. |
| PR alignment | **PASS** | Cleanly merged origin/main (now includes #250/#251/#252/#256). No conflicts despite both PR #256 and this PR modifying `lfmt-infrastructure-stack.ts` — git's text merge resolved them automatically because the SFN IAM (#210) lives in `createIamRoles()` while the CspReport role + route live in distinct sections (lines 1149-1164 + 2192-2222). Verified by full build + test + cdk synth after merge. |

**No Critical / No High findings.** 2 Medium observations (logged, addressed inline or rationalized). 2 Low observations (rationalized).

---

### Critical review by file

#### `backend/infrastructure/lib/csp.ts` — pure CSP builder (#216)

**Strengths**:
- True SRP: 271 lines, no AWS SDK / no CDK imports, unit-testable without `cdk synth`. Reusable from a future Lambda@Edge nonce pipeline for #254.
- `CspDirectives = Partial<Record<CspDirective, string[]>>` — every directive is structurally typed AND narrowed to the canonical name set. Adding a `frame-src` requires extending the union AND the test matrix (forced co-evolution).
- REPLACE-not-MERGE semantics for overrides — the comment at lines 75-79 is correct: a merge mode would require every reader to mentally combine two lists, which was exactly the audit failure that prompted #216.
- `report-uri` validation runs at the top of `buildCsp` (line 161) — fails at `cdk synth`, not at deploy time. Empty-array rejection (line 162) catches the sourceless-directive footgun.

**Red-team check — assertValidCspReportUri (lines 206-271)**:
1. Non-string input → `'must be a string'` throw (defensive against `null`/`undefined`/object coercion via `JSON.parse`).
2. Forbidden chars `[\s;,]` check runs FIRST (line 218) and applies to BOTH concrete URLs AND CDK token strings. A token-laden URL containing `;` is rejected before the token-detection regex runs. ✓
3. `startsWith('https://')` is the protocol gate (line 230) — works for both concrete strings AND `https://${Token[...]}...`. ✓
4. Token escape hatch (line 248) skips the URL-parse step only; the forbidden-char + protocol checks still bite. ✓
5. Pure-function: no `Token.isUnresolved` import → the module is CDK-version-independent.

**Medium observation**: The CDK-token regex `/\$\{Token\[[^\]]+\]\}/` is a lexical contract with CDK internals. A future CDK upgrade that changes the unresolved-reference format would silently bypass the escape hatch and fire `new URL()` on a non-URL string, throwing at synth time. **Mitigation**: CDK guarantees the format will not collide with user data; if the format changes, the existing test (`accepts an unresolved CDK token URL`) breaks visibly at the CDK upgrade PR. Acceptable trade-off for a pure-module architecture.

#### `backend/functions/security/cspReport.ts` — anonymous violation-report Lambda (#201)

**Strengths — defensive design**:
- 7-layer threat model documented in the file header (lines 19-62). Each layer is implemented in code AND has at least one negative test.
- Content-Type normalization (`normalizeContentType`) drops `;charset=` and `;boundary=` parameters AND lowercases — defeats `Application/CSP-Report; charset=utf-8` bypass attempts.
- Body-size cap uses `Buffer.byteLength(rawBody, 'utf8')` (line 321) — char-count `length` would underestimate multibyte payloads. Correct.
- Field allowlist: 9 named fields in the WARN log; any extra/unknown keys in the payload are DROPPED, not preserved as `...rest`. Protects against log-poisoning with forged `userId: "admin"`.
- 400-no-log on malformed body protects against anonymous CloudWatch flooding.
- 204-no-body on success — never echo input, no XS-leak surface.
- Dedicated `CspReportLambdaRole` IAM: basic execution ONLY.

**Red-team check — Reporting API batch handling (lines 202-234)**:
- Non-CSP entries (`e.type !== 'csp-violation'`) silently dropped — honest browser behavior (NEL batches can ride alongside CSP).
- Partially-malformed entries dropped via `continue` — one bad entry doesn't fail the whole batch.
- `effectiveDirective` is the required field per W3C CSP3 Reporting; we fallback to `violatedDirective` for compatibility with browsers emitting the legacy camelCase form (line 217).
- `blockedURL` (modern) ∪ `blockedUri` (legacy) via nullish coalescing (line 224) — handles both Chromium and Firefox shapes.

**Low observation**: If a Reporting API batch contains ONLY non-CSP entries (e.g., a pure NEL batch hitting our endpoint by misconfiguration), the loop drops all entries and we return 400 (records.length === 0). This conflates "no CSP entries" with "malformed". **Not a security issue** — at worst the browser's reporting client logs a 400, which is the correct response for "wrong endpoint". Acceptable.

**Low observation**: `typeof Buffer !== 'undefined'` fallback (line 320) is effectively dead code in both Lambda runtime AND Node-based unit tests. Could be simplified to a direct `Buffer.byteLength(...)`. Stylistic — not worth a fix-up PR.

**Cost ceiling check**: API Gateway throttle at line 1107 of the stack: `100 RPS / 200 burst` (dev), `1000 RPS / 2000 burst` (prod). Worst-case anonymous flood = throttle * 24h * ~1KB per WARN entry ≈ 8.6 GB/day on dev, 86 GB/day on prod. At CloudWatch ingest ~$0.50/GB this is bounded at ~$4.30/day (dev) / ~$43/day (prod). **Bounded by API Gateway, no additional DDB token-bucket required for the POC.** If operational telemetry shows targeted flooding, follow-up could add a DDB token-bucket.

#### `backend/infrastructure/lib/lfmt-infrastructure-stack.ts` — CSP wiring + CspReport route

**Strengths**:
- `CspReportLambdaRole` declared BEFORE `createLambdaFunctions()` runs — the strict-null-check in `createLambdaFunctions()` (line 1565) fails fast if the role isn't created in the expected order.
- Both initial CSP (line 2360) and updated CSP (line 2489) sourced from `buildCsp()` — no drift between the two policies. Pre-#216 the two were independent string-concat blocks.
- `POST /csp-report` route uses `AuthorizationType.NONE` (line 2220) with a comment EXPLICITLY warning that adding Cognito auth here would drop every browser report. Defense against well-intentioned future "harden the API" sweep.
- `corsPreflightOptions('POST')` wired automatically — same-origin XHR from a debug page works without a CORS error.
- `report-uri` directive is emitted in the UPDATED policy (line 2510), NOT the initial — by the time the updated policy is applied, the API Gateway domain is concrete so `assertValidCspReportUri` validates a real URL (token escape hatch is the synth-time pass for the initial deploy).
- `EXPECTED_APPLICATION_LAMBDA_COUNT = 16` annotated with full history (line 1660-1663): `15 → 16 (+1 for CspReport)`.

#### `frontend/e2e/tests/contract/api-envelope-live.spec.ts` + `.github/workflows/e2e-contract-nightly.yml`

**Strengths**:
- Tag-based gating (`@contract-live`) prevents accidental inclusion in the per-PR Playwright run.
- One `beforeAll` login → one Cognito session per file: keeps `adminInitiateAuth` burst rate well under the dev pool limit.
- `loginAndGetIdToken` uses **idToken** (not accessToken) as the Bearer credential — matches the API Gateway CognitoUserPoolsAuthorizer behavior. The function's docstring explicitly references the authorizer config in the stack.
- Per-test wire-shape regression guards: `expect(body).not.toHaveProperty('uploadUrl')` (line 230) catches a flat-vs-wrapped envelope migration; `expect(body).not.toHaveProperty('chunksTranslated')` (line 275) catches a #229 field-rename regression.
- Workflow uses `concurrency: e2e-contract-nightly` (line 44) — nightly + a hot-fix `workflow_dispatch` don't race on the shared Cognito session.
- Fork-safe: secrets fall back to per-run user via Cognito auto-confirm — forked PRs running `workflow_dispatch` still work without secret access.

**Medium observation**: Per-run fallback writes test users to dev Cognito. After ~365 nightly runs (or one year), the pool accumulates ~365 zombie users (assuming the shared secrets path is taken — which it should be in normal operation). **Mitigation**: when shared secrets are configured (the default operational mode), the fallback is never hit. The accumulation is bounded by how often secrets are absent. Tracking as a Low follow-up for a Cognito-GC sweep is appropriate but not blocking.

---

### Issue split rationale (#197 → #254 + #255) — verification

PR splits the original #197 (style-src nonces + httpOnly cookies) into two follow-ups based on:
- **Different architectural layer**: #254 nonces require Lambda@Edge (CloudFront-tier), out-of-scope per #197 issue body.
- **Different concern class**: #255 httpOnly cookies are auth architecture, not CSP.
- **Cross-domain topology blocker**: SameSite=Lax cookies don't reliably cross `cloudfront.net` → `execute-api.amazonaws.com`. Real blocker, not a punt.
- **CSRF mitigation choice**: owner design decision required before #255 implementation.

**Confirmed**: both #254 and #255 were filed with full justifications. The split is appropriate.

---

### Items explicitly OUT of scope for this PR

| Item | Why deferred |
|---|---|
| Lambda@Edge style-src nonces | #254 (split). Requires Lambda@Edge pipeline + frontend/backend deploy sync. Documented as future work in `csp.ts:104-115`. |
| httpOnly cookies for auth | #255 (split). Requires custom-domain ACM/Route53 + CSRF mitigation choice. |
| DDB token-bucket per-IP rate limit on `/csp-report` | YAGNI for the POC. API Gateway throttle bounds the worst case at ~$43/day on prod; the DDB write would cost more in steady state than the protection it offers. |
| Cognito GC for fallback-path test users | Operations sweep; not blocking. Only relevant when shared CI secrets are absent. |

---

### Validation summary

| Package | Build | Lint | Format | Tests |
|---|---|---|---|---|
| `shared-types` | **PASS** | — | — | — |
| `backend/functions` | **PASS** | **PASS** | **PASS** | **598/598 PASS** (3 pre-existing skips) |
| `backend/infrastructure` | **PASS** | — | — | **89/89 PASS** + `cdk synth --context environment=dev` **PASS** |

**Test count delta**: 598 functions tests (+25 vs base for cspReport.test.ts) / 89 infra tests (+12 vs base for csp.ts builder + assertValidCspReportUri + report-uri assembly).

---

### Diff summary

```
.github/workflows/e2e-contract-nightly.yml                             | 119 ++ (NEW workflow — nightly + dispatch)
backend/functions/security/cspReport.test.ts                          | 400 ++ (NEW — 25 handler tests)
backend/functions/security/cspReport.ts                               | 372 ++ (NEW — anonymous Lambda)
backend/infrastructure/lib/__tests__/infrastructure.test.ts           | 216 ++ (CSP + lambda count drift guard)
backend/infrastructure/lib/csp.ts                                     | 271 ++ (NEW — pure CSP builder)
backend/infrastructure/lib/lfmt-infrastructure-stack.ts               | 296 ++ (Role + route + 2× buildCsp call sites)
frontend/LOCAL-TESTING.md                                              |  40 ++ (mock-mode usage doc)
frontend/e2e/tests/contract/api-envelope-live.spec.ts                  | 407 ++ (NEW — @contract-live suite)
8 files changed, 1919 insertions(+), 202 deletions(-)
```

---

**Ready for review.** No blockers, no follow-up issues required (medium/low observations are documented with rationale or already filed as #254/#255).

leixiaoyu:
## OMC R2 Self-Review — Wave 2 Track D (self-review-findings resolution)

**Triggered by**: OMC R1 self-review on this PR ([comment](https://github.com/leixiaoyu/lfmt-poc/pull/257#issuecomment-4440710613)) — user explicitly requested that R1 findings be resolved rather than rationalized.
**Reviewed commit**: this push (OMC R2 fixes) against `e3dfe82` (R1 reviewed commit).
**R1 findings addressed**: 2 Medium + 2 Low (all four).
**New tests added**: +5 (1 CDK-upgrade drift guard + 4 Reporting-API semantic split).
**Date**: 2026-05-13

---

### R1 finding → R2 resolution map

| R1 finding | Severity | R1 conclusion | R2 resolution |
|---|---|---|---|
| CDK-token regex coupling | Medium | "Acceptable trade-off" | **In-PR drift guard test** — `infrastructure.test.ts` now generates a real CDK token via `Lazy.uncachedString()` + `Token.isUnresolved()` and asserts the csp.ts regex matches CDK's actual lexical output. A future CDK upgrade that changes the `${Token[<id>]}` format breaks this test at the upgrade PR, not at next deploy. |
| Per-run user accumulation in dev Cognito | Medium | "Tracking as a Low follow-up… not blocking" | **New weekly cleanup workflow** — `.github/workflows/cognito-test-user-cleanup.yml`. Sunday 10:00 UTC schedule + `workflow_dispatch` for on-call. OIDC auth via existing `secrets.AWS_ROLE_ARN`. Defenses-in-depth: email-prefix filter + domain double-check + 7-day age filter + dry-run mode + per-user job-log audit trail. |
| NEL-only Reporting-API batch returned 400 | Low | "Not a security issue. Acceptable." | **Semantic split** — `parseReportsApiBody` return type widened to `SanitizedCspReport[] \| null`. `null` = structurally malformed (not a JSON array) → 400. `[]` = valid array with zero CSP entries → 204 (no log). `[r,...]` = log each + 204. Legacy `application/csp-report` parser keeps `null` for missing wrapper/directive (no "valid-but-empty" state is possible for the single-record shape). |
| `Buffer.byteLength` fallback dead code | Low | "Stylistic — not worth a fix-up PR." | **In-PR removal** — `Buffer` is a Node builtin always present in both Lambda runtime AND Jest test runner. The `typeof Buffer !== 'undefined'` ternary is unreachable. Direct `Buffer.byteLength(rawBody, 'utf8')` call now. |

---

### Critical review of each fix (red-team posture)

#### 1. CDK-token drift guard (Medium #1)

**File**: `backend/infrastructure/lib/__tests__/infrastructure.test.ts:2210-2253`

**What it does**: Uses `Lazy.uncachedString({ produce: () => 'resolved' })` to obtain a real CDK Token instance, verifies `Token.isUnresolved()` on it (sanity check on the test setup itself), stringifies it to capture CDK's actual lexical format, then asserts the inline regex `/\$\{Token\[[^\]]+\]\}/` matches AND `assertValidCspReportUri` accepts the full URL.

**Red-team — is the test setup self-validating?**
Yes. The `expect(Token.isUnresolved(lazyToken)).toBe(true)` line catches the case where CDK changes how `Lazy.uncachedString` works (test infrastructure breaks visibly, not silently passes).

**Red-team — does the regex live in two places now?**
Yes — inline in csp.ts:248 AND inline in this test. That's intentional. csp.ts is a pure module that does NOT export the regex (preserving the minimal public API surface — only `buildCsp` + `assertValidCspReportUri` are exported). The test comment at lines 2237-2241 explicitly calls this out and instructs future contributors who change csp.ts to also update this test. This is a deliberate "two-sided drift guard" — it catches CDK format changes (the original Medium-1 concern) AND catches csp.ts regex changes (the inverse drift).

**Red-team — could a CDK version that drops `Lazy.uncachedString` cause a false positive?**
Possible but cheap to fix — the API has been stable across CDK v1/v2. If a future CDK removes it, the test compile fails at the import — visible failure mode.

**Verdict**: ✅ Drift guard is load-bearing; the previous "acceptable trade-off" rationalization is no longer needed.

#### 2. Cognito cleanup workflow (Medium #2)

**File**: `.github/workflows/cognito-test-user-cleanup.yml` (NEW, 187 lines).

**Defenses-in-depth (5 independent layers)**:
1. **Prefix filter** at Cognito ListUsers (`email ^= "contract-"`)
2. **Domain double-check** in shell (`*@e2e-test.lfmt.com`) — survives a future spec rename that uses the same prefix in a different domain
3. **Age filter** (>7 days) — operations SLA for nightly-failure triage
4. **Dry-run mode** for `workflow_dispatch` — operators can preview before destroying
5. **Per-user audit trail** in job log — every `DELETE` / `PRESERVE` / `SKIP` decision printed

**Red-team — could the workflow delete a non-test user?**
Three filters would all have to misfire simultaneously. The cross-product of (prefix-match) ∧ (domain-match) ∧ (age >7d) is extremely narrow.

**Red-team — could the OIDC role lack the required IAM?**
Possible. The workflow runs `aws sts get-caller-identity` early; a missing trust policy fails loudly with a clear error message. The workflow header documents the IAM dependency. If the existing deploy role doesn't grant `cognito-idp:AdminDeleteUser`, the first run on Sunday fails visibly and that becomes a 5-min IAM update PR. **Acceptable failure mode** for a new operational workflow.

**Red-team — schedule + OIDC trust policy interaction**
GitHub OIDC trust policies typically allow assumption based on `sub: repo:<org>/<repo>:ref:refs/heads/<branch>`. Scheduled workflows run with `sub: repo:<org>/<repo>:ref:refs/heads/main` (same as a push to main). The existing deploy role accepts main-branch OIDC tokens, so the trust path is the same.

**Red-team — concurrency**
The `concurrency: cognito-test-user-cleanup / cancel-in-progress: false` group serializes Sunday-scheduled + a manual dispatch on the same day. The `cancel-in-progress: false` is deliberate: a manual run that arrives while Sunday is still processing should QUEUE behind it (not cancel) so the long-running pagination loop can finish.

**Red-team — bash portability**
The shell uses `date -u -d ... +%s` (GNU date), with fallbacks to `gdate` (macOS via coreutils) and finally to a Python fromisoformat one-liner. GitHub Actions `ubuntu-latest` runners have GNU date, so the primary path always works on CI. The fallbacks exist for local dry-run-via-act testing.

**Verdict**: ✅ All five defense layers are independent and would have to fail simultaneously for a wrong-user-deleted outcome.

#### 3. Reporting-API semantic split (Low #1)

**Files**: `backend/functions/security/cspReport.ts:202-234` (parser), lines 337-378 (handler dispatch); tests at `cspReport.test.ts:196-225` + `262-322`.

**Red-team — does the new 204-on-empty-batch open a new attack surface?**
An attacker can now send `[]` and receive 204 instead of 400. But:
- **Zero log emission** — CloudWatch cost ceiling is unchanged. The flood-prevention guarantee (no per-request log on validation-failure path) is preserved.
- **Throttle bound still applies** — API Gateway 100 RPS (dev) / 1000 RPS (prod) caps the abuse rate.
- **No state mutation** — endpoint is read-only beyond CloudWatch.
- **204 doesn't echo input** — XS-leak surface unchanged.

The trade-off: legitimate non-CSP batches (browser misconfig) now get a "correct" 204 rather than a misleading 400. **Net: behavior more correct, cost unchanged.**

**Red-team — does the legacy `application/csp-report` path lose its 400-on-malformed behavior?**
No. `parseLegacyReport` still returns `null` when the wrapper is missing or `violated-directive` is absent (lines 168-176). The handler dispatch (lines 339-350) explicitly wraps the legacy parser result with `return one ? [one] : null` — null propagates to the 400 branch. Verified by the existing test `'returns 400 (no log) when LEGACY payload contains zero valid reports'`.

**Red-team — return-type change is API-breaking for `parseReportsApiBody` consumers**
The function is **exported only for unit testing** per its docstring (line 200). No production caller exists outside the same file. The handler is the sole consumer and was updated in lockstep. The test file was updated to match (5 existing assertions migrated; 4 new test cases added).

**Red-team — could the new dispatch logic regress an existing test?**
Ran `npm test` on backend/functions: **602 tests pass** (was 598 at R1 baseline; +4 for the new differentiated cases). The 4 pre-existing reporting-api tests (parser + handler) were updated mechanically to match the new return contract.

**Verdict**: ✅ Behavior is more semantically correct without compromising any defense layer.

#### 4. Buffer fallback removal (Low #2)

**File**: `backend/functions/security/cspReport.ts:316-322`.

**Red-team — is there ANY environment where `Buffer` is undefined that this handler could run in?**
- Lambda runtime (Node.js 22.x): Buffer is a global. Always present.
- Jest test runner (Node-based): Buffer is a global. Always present.
- A future "edge" rewrite to Cloudflare Workers / Deno would lack Buffer — but that's an architectural change, not a runtime variance. The comment in the file (lines 316-319) now states the assumption explicitly so a future port can't miss it.

**Verdict**: ✅ Dead code removed; assumption documented.

---

### Items NOT addressed (and why)

| Item | Why not |
|---|---|
| Cost-ceiling alarm for `/csp-report` CloudWatch ingest | YAGNI. The 100 RPS throttle (dev) bounds worst-case ingest at ~$4.30/day; operations alarm-budget is on the order of dollars not pennies. If observed traffic justifies a finer-grained alarm, that's a separate PR with a CloudWatch metric filter. |
| Per-request DDB token-bucket rate limit on `/csp-report` | Same. The API Gateway throttle is the right layer for flood control on an anonymous endpoint; adding DDB writes per report would multiply the cost-per-report by ~100x. |
| Promoting the CDK-token regex to a constant in csp.ts (and an export) | Would expand the public API surface of a deliberately-minimal pure module. The R1 finding was about the regex's coupling to CDK internals, not its location. The drift guard test is the correct fix; an export would just relocate the coupling. |

---

### Validation

| Package | Build | Lint | Format | Tests |
|---|---|---|---|---|
| `shared-types` | **PASS** | — | — | — |
| `backend/functions` | **PASS** | **PASS** | **PASS** | **602/602 PASS** (3 pre-existing skips; +4 vs R1 for the Reporting-API semantic split) |
| `backend/infrastructure` | **PASS** | — | — | **90/90 PASS** (+1 vs R1 for the CDK-token drift guard) + `cdk synth --context environment=dev` **PASS** |
| `.github/workflows` | — | — | — | `actionlint` **CLEAN** on new workflow; existing path-filter test still **PASS** |

**Test delta vs R1**: +5 (4 cspReport + 1 infrastructure).

---

### Diff summary

```
.github/workflows/cognito-test-user-cleanup.yml       | 187 ++ (NEW — weekly Cognito GC, dry-run + on-call dispatch)
backend/functions/security/cspReport.ts               |  73 ++- (parseReportsApiBody return-type widened + handler dispatch split + Buffer dead-code removal)
backend/functions/security/cspReport.test.ts          |  92 ++- (5 existing assertions migrated to null-check + 4 new cases for the semantic split)
backend/infrastructure/lib/__tests__/infrastructure.test.ts | 47 ++ (CDK-token drift guard test + new Lazy/Token imports)
4 files changed, 363 insertions(+), 30 deletions(-)
```

---

**Ready for re-review.** All four R1 findings resolved with in-PR changes — no follow-up issues filed.

xlei-raymond (approved):
As a Senior Reviewer, I have conducted an exhaustive audit of PR #257. This PR is a tour-de-force in Security Engineering and Operational Maturity. By hardening our CSP infrastructure, establishing a telemetry loop for violations, and introducing a
  live-backend contract guard, the team has significantly reduced our production risk profile just before the investor demo.

  ---

  1. Architectural Excellence: Pure-Function Infrastructure
  The extraction of backend/infrastructure/lib/csp.ts into a pure, testable module is an exemplary architectural win.

  Educational Note:
  Infrastructure code is notoriously hard to unit-test because it often requires a heavy framework (CDK) and a mock stack.
   * The Lesson: Decouple Logic from Constructs. By making buildCsp a pure function that only deals with strings and arrays, the team has enabled a massive test matrix (25+ cases) that runs in milliseconds without synthesising a single CloudFormation resource.
     This is how you build reliable infrastructure-as-code.

  2. Defense in Depth: The cspReport Lambda
  The design of the POST /csp-report handler (Issue #201) is a masterclass in Defensive Programming for Internet-Facing Endpoints.

  Mentorship Note:
  Unauthenticated endpoints are "Front Doors" for attackers. A junior approach would be to simply log the JSON.parse result.
   * The Lesson: Trust No One. The team implemented a multi-layered shield:
       1. Ingress Guard: Content-Type allowlist and 64KB body cap before parsing.
       2. Logic Guard: Discrimination between legacy and Reporting-API shapes (OMC R2 Low-1 split).
       3. Egress Guard: A strict field allowlist and 2KB truncation on log emission.
      This prevents log-poisoning, CloudWatch cost-bloating, and anonymous log-flooding. This is the "Gold Standard" for unauthenticated telemetry handlers.

  3. Reliability: The "Live Contract" Guard
  The introduction of the api-envelope-live.spec.ts (Issue #219) closes a critical gap in our CI/CD philosophy.

  Critique:
  Unit tests (and even MSW-based integration tests) are "Mocks talking to Mocks." They cannot catch drift in the deployed environment (e.g., a manual AWS console change or a shared-types version mismatch in the Lambda bundle).
   * The Lesson: Anchor your types to Reality. By hitting the actual deployed dev environment nightly, we ensure that our TypeScript interfaces aren't just "good ideas"—they are verified contracts. The tag @contract-live and the dedicated nightly workflow are
     perfect for managing the quota and flakiness risks associated with live tests.

  4. Operational Maturity: Garbage Collection
  The addition of .github/workflows/cognito-test-user-cleanup.yml (OMC R2 Medium-2) demonstrates high professional accountability.

  Observation:
  Identifying that the live-contract spec creates "zombie" users and proactively automating their cleanup before they bloat the Cognito pool shows that the team is thinking about Long-Term System Health, not just the next feature.

  5. Pro Move: The CDK-Token Drift Guard
  I was particularly impressed by the test in infrastructure.test.ts that uses Lazy.uncachedString() to falsify the CDK-token regex.

  Educational Note:
  Our csp.ts module uses a lexical regex to allow CDK tokens through its URL validator. This creates a hidden coupling: if a future CDK upgrade changes the ${Token[...]} format, our build would silently start failing.
   * The Lesson: Explicitly verify your assumptions. By generating a real CDK token and asserting the regex matches it, the team has created a "Drift Alarm" that will fire at the time of a CDK upgrade PR, rather than at a random deploy time.

  ---

  Final Verdict:
  APPROVE.

  This is an exceptionally high-quality PR. It demonstrates a deep understanding of web security, distributed system reliability, and operational hygiene. The thorough self-review and the resolution of all OMC findings in-PR show a level of discipline that is
  truly commendable.

  Team: Excellent work. The buildCsp refactor makes #197/nonces a trivial follow-up. This is the correct foundation for our production launch.

  Merge immediately.

Issue #240, opened 2026-05-10
infra(ci): deploy-frontend.yml E2E job has no timeout-minutes — hung run pinned the deploy concurrency lock for 40+ min


leixiaoyu:
## Summary

On 2026-05-10, the \`Run E2E Tests\` step inside \`deploy-frontend.yml\` (run [25638423405](https://github.com/leixiaoyu/lfmt-poc/actions/runs/25638423405)) hung 40+ minutes after the deploy step had already succeeded. Because the workflow uses \`concurrency: deploy-\${{ github.ref }}\` with \`cancel-in-progress: false\`, the hung run pinned the lock and blocked PR #239's backend AND frontend deploys from starting.

Specifically:
- 20:02 — PR #238 merge → frontend run starts
- 20:05 — Deploy Frontend to Development job succeeds (code live on CDN)
- 20:06 — Run E2E Tests step starts… and hangs
- 20:13 — PR #239 merges. Backend deploy queues behind the lock; frontend deploy is auto-cancelled by GitHub's queue-dedup at creation
- 20:46 — manual recovery: cancel the hung run to release the lock

**Net effect**: 40 minutes of demo-prep verification time lost; PR #239's frontend bits never deployed automatically (cancelled in queue, will not auto-restart).

## Root cause

\`deploy-frontend.yml\` does not set \`timeout-minutes\` on the \`e2e-tests\` job. GitHub Actions defaults to a 6-hour budget — far longer than any sensible Playwright run. A flake or stuck browser binds the lock for hours.

## Proposed fix

**Option A** (minimal): add \`timeout-minutes: 25\` to the \`e2e-tests\` job in \`.github/workflows/deploy-frontend.yml\`. A normal post-deploy E2E run takes ~5–10 minutes; 25 gives generous headroom.

**Option B** (architecturally cleaner): remove the \`e2e-tests\` job from the \`deploy-\${{ github.ref }}\` concurrency group entirely, OR move it to a separate workflow that doesn't gate the next deploy. E2E against deployed-dev is *verification*, not part of deploy itself; failing/hung E2E should not prevent the next deploy from starting.

**Option C** (defensive): set \`timeout-minutes\` AND keep the concurrency group. Belt-and-braces.

Recommend Option C — simplest delta + no architectural debate, fixes the immediate symptom and the recurrence risk in one PR.

## Additional follow-up

[...]

Issue #171, opened 2026-05-02
[BUG-LOW] capture-chapter-metrics.mjs default LFMT_TEST_EMAIL hardcoded to '2026-04-25' (docstring claims 'fresh per-day')


leixiaoyu:
## Severity
LOW — script defect, easy workaround (set \`LFMT_TEST_EMAIL\` env var explicitly).

## Symptom

\`demo/scripts/capture-chapter-metrics.mjs:74\`:

\`\`\`javascript
const TEST_EMAIL = process.env.LFMT_TEST_EMAIL || 'claude-track-b-2026-04-25@lfmt-poc.dev';
\`\`\`

Docstring at line 32 claims:
\`\`\`
LFMT_TEST_EMAIL — Optional. Defaults to a fresh per-day address.
\`\`\`

But the literal default is the date of the original failed 2026-04-25 capture attempt, not "today." Re-running the script without setting \`LFMT_TEST_EMAIL\` reuses the original (orphaned) Cognito account, whose password (\`TrackBDemo!2026\`) was the burned credential removed by PR #146's OMC fix R7.

## Reproduction

1. Set \`TEST_PASSWORD\` to anything other than the original \`TrackBDemo!2026\`.
2. Run \`node demo/scripts/capture-chapter-metrics.mjs\` (no \`LFMT_TEST_EMAIL\`).
3. Login attempt fails with 401 → register attempt fails with 409 (user exists) → script aborts.

## Suggested fix

Replace the literal default with a computed one:

\`\`\`javascript
const TODAY = new Date().toISOString().slice(0, 10); // '2026-05-02'
const TEST_EMAIL = process.env.LFMT_TEST_EMAIL || \`claude-track-b-\${TODAY}@lfmt-poc.dev\`;
\`\`\`

This matches the docstring's stated behavior. Each day's run gets a fresh account; same-day re-runs reuse the per-day account (matches the existing same-day RPD guard's intent).

**Side benefit**: orphan Cognito test accounts only accumulate at 1/day instead of 1/run.

## Related

- PR #146 (introduced the script + the docstring/literal mismatch)
- OMC review's "account-cleanup tooling" suggestion (deferred during PR #146 implementation)

PR #258, opened 2026-05-13
chore(infra): staging/prod deploy-pipeline parity sweep — bootstrap guard + diff-short-circuit decision (#253)

### LINKED ISSUE #253: chore(infra): staging/prod deploy-pipeline environment-parity sweep
## Problem

After the #247/#248 fold-in (PR #250), two deploy-pipeline consistency gaps remain in `deploy-backend.yml`:

### Gap 1 — staging missing `cdk bootstrap` guard (pre-existing, touched-adjacent in #250)

`deploy-dev` (line 224-233) and `deploy-prod` (line 663-672) both have an idempotent CDKToolkit check:
\`\`\`bash
if ! aws cloudformation describe-stacks --stack-name CDKToolkit ...; then
  npx cdk bootstrap ...
fi
\`\`\`
`deploy-staging` has **no** such guard. First-ever deploy of staging against a fresh account/region will fail with `This stack uses assets, so the toolkit stack must be deployed to the environment (Run "cdk bootstrap aws://..."`.

**File boundary:** `deploy-backend.yml` lines 583-585 (`CDK Deploy to Staging` has no preceding bootstrap step).
**Concern class:** deploy-pipeline architecture — operational correctness for fresh environments.

### Gap 2 — `#163` diff short-circuit is dev-only; staging/prod deploy unconditionally

`deploy-dev` uses a `cdk diff` short-circuit (lines 255-269) that skips CDK deploy when there are no infrastructure changes, saving ~5-8 min per no-op. `deploy-staging` and `deploy-prod` call `npx cdk deploy` unconditionally.

This was intentional in Issue #163 (scoped to dev). The question for staging/prod is whether "skip deploy if no diff" is correct behaviour there:
- **Argument for keeping unconditional:** staging/prod deploys are manual (`workflow_dispatch`), implying intent to deploy; skipping on no-diff could surprise operators.
- **Argument for adding short-circuit:** a staging deploy triggered accidentally (or as a smoke-check) still takes 5-8 min; the short-circuit would make it ~30s.

prod has a `cdk diff || true` audit step (line 674-676) but no has_changes gate — it always proceeds to `cdk deploy`. Staging has neither.

**File boundary:** `deploy-backend.yml` lines 583-585 (staging), 678-680 (prod).
**Concern class:** deploy-pipeline architecture — design decision, not a straightforward CI hygiene fix. Needs explicit owner decision before implementation.

## Why deferred from PR #250

PR #250 scope is CI/CD pipeline hygiene (Issues #160/#163/#187/#247/#248/#249). Folding in the bootstrap guard is mechanical; folding in the short-circuit design decision is not. Adding both inline without a design decision risks introducing a silent behaviour change to staging/prod that operators rely on (always-runs = auditable intent signal).

## Recommended action

1. Add `cdk bootstrap` guard to `deploy-staging` (mechanical, low-risk).
2. Decide: should staging/prod get the `#163` short-circuit? If yes, mirror `deploy-dev`'s pattern with an explicit `has_changes` gate and document that manual dispatches are still advisory (not guaranteed to skip). If no, add a comment in `deploy-staging` / `deploy-prod` documenting the intentional asymmetry.

leixiaoyu:
## Summary

Closes #253. Pipeline-consistency sweep on `.github/workflows/deploy-backend.yml`. Originally scoped to two gaps from the issue body; expanded in a second commit to fold in three additional out-of-scope items at owner request, where each fold-in was mechanical and self-contained. The fourth originally out-of-scope item is genuinely architectural and deferred to #260.

**Diff size**: 1 file changed, 122 insertions (+) across two commits.

## Gap 1 — staging missing `cdk bootstrap` guard (mechanical fix, commit 1a84bc4)

**Problem**: `deploy-dev` and `deploy-prod` already have an idempotent `CDKToolkit` check; `deploy-staging` did not. First-ever deploy of staging against a fresh account/region would fail with `This stack uses assets, so the toolkit stack must be deployed to the environment`.

**Fix**: Added the same `if ! aws cloudformation describe-stacks --stack-name CDKToolkit ...` block immediately before the staging `cdk deploy` step. The block is **textually identical** to the dev and prod versions for grep-ability — operators triaging a fresh-account bootstrap can search the same string across all three jobs.

## Gap 2 — `#163` diff short-circuit is dev-only (design decision: keep unconditional on staging/prod, commit 1a84bc4)

**Problem statement**: `deploy-dev` runs a `cdk diff` short-circuit (Issue #163) that skips `cdk deploy` when no infrastructure changes are detected, saving ~10 min per noop deploy. `deploy-staging` and `deploy-prod` deploy unconditionally. The issue body flagged this as a design question, not a clear bug — strong arguments existed both ways.

**Decision**: **keep staging/prod unconditional.** Add documenting comment blocks on both the staging and prod `cdk deploy` steps so a future contributor reading the asymmetry doesn't "fix" it by reflex.

**Reasoning** (in priority order):

1. **Trigger semantics differ**. `deploy-dev` fires on every push to `main` AND on `workflow_dispatch`. The push-triggered path runs on shared-types-only or backend-only commits that may have zero infra impact — that's exactly the noop case #163 optimizes away. `deploy-staging` and `deploy-prod` are **`workflow_dispatch` ONLY** (job-level `if:` guard explicitly excludes the push event). Every staging/prod invocation carries explicit operator intent to deploy.
2. **UX of a manual dispatch exiting in ~30s with "no changes"** is ambiguous: did the deploy actually run? Was the frontend rebuild step gated out? Did the smoke check validate the new code? Operators dispatching a staging/prod deploy expect a full deploy, not a conditional fast-path. The cost savings (5-8 min on a noop) doesn't outweigh the operator-confidence cost.
3. **Audit-trail-first design**. `deploy-prod` already has a `CDK Diff (show changes for audit trail)` step that records the proposed change-set in the workflow log BEFORE the deploy runs. Adding a `has_changes` gate would either (a) duplicate that diff invocation (wasted ~30s) or (b) replace it (lost paper trail). Both are strictly worse than the current always-deploy behaviour.

The inline comment block on both staging and prod explicitly enumerates these three points so this decision is grep-able and self-explaining when the workflow is read in isolation.

## Expansion round — three additional parity gaps folded in (commit e284b69)

Owner reviewed the original PR + report and asked for the previously out-of-scope items to be folded in where each was mechanical and self-contained. Three folded in; one deferred (see "Deliberately deferred" below).

### Gap 3 — `timeout-minutes` missing on all three deploy jobs

**Problem**: `smoke-tests` (15min) and `integration-tests` (30min) got timeouts in PR #240 to keep a hung run from holding the `deploy-${ref}` concurrency lock indefinitely. The three actual deploy jobs (`deploy-dev`, `deploy-staging`, `deploy-prod`) were skipped in that pass — a hung `cdk deploy` against a stuck CloudFormation update would hold the lock for the GitHub Actions 6-hour cap.

**Fix**: Added `timeout-minutes` to all three:

- `deploy-dev`: `30` (line 152). Observed noop run 25770555446 finished in 3m48s; a real deploy with frontend rebuild lands inside 15. 30 gives ~6x headroom.
- `deploy-staging`: `30` (line 533). Full-deploy path (no diff short-circuit per Gap 2). ~5-8 min CDK + ~2-3 min frontend + ~3 min CloudFront. 30 gives ~2x headroom.
- `deploy-prod`: `45` (line 702). Extra room for the additional `CDK Diff (audit trail)` step and Cognito user-pool updates which can take 5-10 min on their own.

### Gap 4 — `Verify AWS identity` step missing from `deploy-staging`

**Problem**: Present on `deploy-dev` (line 209) and `deploy-prod` (line 736), missing on `deploy-staging`. Without it, a misconfigured OIDC trust that silently lands the workflow in the wrong account has no audit signal until something downstream breaks.

**Fix**: Added the same `aws sts get-caller-identity` step on staging immediately after the `Configure AWS credentials (OIDC)` step (line 587). 4-line copy/paste; ~200ms operational cost.

### Gap 5 — No post-deploy API health check on `deploy-staging` / `deploy-prod`

**Problem**: `deploy-dev` has a `Test API Health` curl step (line 312) that catches the CFN-succeeds-but-API-unreachable failure mode (DNS lag, custom-domain misconfig, broken Lambda authorizer). Staging and prod had nothing equivalent — a deploy that succeeded at the CFN layer but produced an unreachable API would only be caught by the operator's manual smoke.

**Fix**: Mirrored the dev curl on both staging (line 683) and prod (line 808). The composite `rebuild-frontend` action already exports `steps.frontend.outputs.api_url`, so adapting was a single-line URL swap. The curl is non-blocking (`|| echo …`) by design — a transient 5xx during custom-domain propagation should not fail a manual staging/prod dispatch, but the line in the run log is the audit signal that lets operators re-check if the manual smoke later reveals trouble.

## Validation

| Check | Result |
|---|---|
| `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/deploy-backend.yml'))"` | **PASS** |
| `python3 tests/workflows/test_deploy_path_filters.py` | **PASS** (`All deploy workflow path-filter contracts satisfied.`) |
| `/tmp/actionlint .github/workflows/deploy-backend.yml` | **40 warnings before, 40 after** — zero new warning classes. The new curl steps use `${{ steps.frontend.outputs.api_url }}` GitHub-expression interpolation (not shell expansion), so they don't add SC2086 warnings. The staging bootstrap SC2086 from commit 1a84bc4 is unchanged (copy-fidelity with dev/prod per #253 brief). |

### Item 4 (original) — `smoke-tests` / `integration-tests` only run after `deploy-dev`

**Why deferred**: This is **architectural work, not a hygiene fix**.

The smoke test (`backend/tests/smoke/smoke.test.ts`) and integration tests (`backend/functions/**/*.integration.test.ts`) consume `secrets.DEV_USER_POOL_ID` and the dev OIDC role + dev Cognito pool. Routing equivalent credentials per environment touches:

1. **GitHub Secrets routing**: needs `STAGING_USER_POOL_ID`, `PROD_USER_POOL_ID` (or environment-protection-scoped secrets, which is the GitHub-native answer).
2. **IAM trust**: OIDC role's trust policy needs broadening, OR per-env roles need minting.
3. **Cognito test-user isolation**: the per-day test-email pattern (#171) is currently dev-only and would need to be replicated for staging/prod with cleanup.
4. **Failure-mode design**: staging smoke against a prod-like pool would create real-looking user rows; needs a policy on whether these tests are isolated, idempotent, or destructive.

Folding into this PR would significantly grow scope and introduce new failure modes (broken IAM, leaked test users in staging/prod) — exactly the criteria the owner gave for declining.

**Tracking**: [#260](https://github.com/leixiaoyu/lfmt-poc/issues/260) — `ci(deploy): add post-deploy smoke + integration tests to staging/prod`.

## Test plan

- [x] `python3 tests/workflows/test_deploy_path_filters.py` — passes (both commits)
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` — passes (both commits)
- [x] `/tmp/actionlint .github/workflows/deploy-backend.yml` — no new warning classes after expansion-round commit (verified: 40 warnings on the pre-expansion stash vs 40 warnings on the post-expansion working tree)
- [x] Mental-model trace of the bootstrap guard: fresh account (describe-stacks fails → bootstrap runs) and existing account (describe-stacks succeeds → bootstrap skipped) — both paths produce the desired behaviour
- [x] Mental-model trace of Gap 2 decision against the issue body's pro/con list — operator-dispatch intent + audit-trail-first design outweigh the ~10min cost savings on the noop case
- [x] Diff review of expansion commit: 5 distinct additive blocks, each `Boy-scout (#253 expansion round)` annotated for grep-ability
- [ ] Owner review of the Gap 2 decision (this remains the load-bearing review item) and the expansion fold-in choices

### DIFF
--- a/.github/workflows/deploy-backend.yml
+++ b/.github/workflows/deploy-backend.yml
@@ -148,6 +148,16 @@ jobs:
     name: Deploy Backend to Development
     runs-on: ubuntu-latest
     needs: test
+    # Boy-scout (#253 expansion round): mirror the `timeout-minutes` discipline
+    # that `smoke-tests` (15) and `integration-tests` (30) already enforce
+    # (added by PR #240). A stuck `cdk deploy` against a hung CloudFormation
+    # update would otherwise hold the `deploy-${ref}` concurrency lock
+    # indefinitely, blocking every subsequent deploy + the parallel
+    # deploy-frontend.yml workflow (shared concurrency group). Observed noop
+    # deploy-dev runs finish in <5 min (e.g. run 25770555446: 3m48s); a real
+    # deploy with frontend rebuild + CloudFront invalidation lands well inside
+    # 15 min. 30 gives ~6x headroom while still failing loud if CFN hangs.
+    timeout-minutes: 30
     # Trigger conditions:
     #   - push to main → automatic deploy
     #   - workflow_dispatch with environment=dev → manual deploy (escape hatch)
@@ -520,6 +530,13 @@ jobs:
     name: Deploy Backend to Staging
     runs-on: ubuntu-latest
     needs: test
+    # Boy-scout (#253 expansion round): same `timeout-minutes` discipline as
+    # deploy-dev above. Staging deploys are always full (no diff short-circuit
+    # — see Gap 2 rationale block below `Configure AWS credentials`) so the
+    # expected wall-clock is closer to dev's full-deploy path: ~5-8 min CDK
+    # + ~2-3 min frontend rebuild + ~3 min CloudFront wait. 30 gives ~2x
+    # headroom and still trips before the GitHub Actions 6-hour cap would.
+    timeout-minutes: 30
     # Issue #160 (branch-guard asymmetry): previously only checked environment
     # input but not github.ref, allowing any branch to deploy to staging via
     # `gh workflow run deploy-backend.yml --ref feature/foo -f environment=staging`.
@@ -566,6 +583,14 @@ jobs:
           role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
           role-session-name: GitHubActions-LFMT-Deploy-Staging
           aws-region: ${{ env.AWS_REGION }}
+
+      # Boy-scout (#253 expansion round): mirror deploy-dev (line 209) and
+      # deploy-prod (analogous step). Surfacing the assumed-role ARN in the
+      # workflow log is part of the audit trail for any staging deploy and
+      # gives operators a fast triage signal when an OIDC misconfiguration
+      # silently lands the workflow in the wrong account.
+      - name: Verify AWS identity
+        run: aws sts get-caller-identity
 
       - name: Install shared-types dependencies
         working-directory: shared-types
@@ -648,6 +673,19 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
           node-version: ${{ env.NODE_VERSION }}
 
+      # Boy-scout (#253 expansion round): mirror deploy-dev's `Test API Health`
+      # step (line 312). Without this, a deploy that succeeded at the CFN
+      # layer but produced an unreachable API Gateway (DNS lag, custom-domain
+      # misconfig, broken Lambda authorizer) would only be caught by the
+      # operator's manual smoke. The curl is non-blocking (`|| echo …`)
+      # because API Gateway can take a few seconds post-deploy to propagate
+      # custom-domain mappings — a transient 5xx is not a deploy failure,
+      # but the line in the run log is the audit signal that the operator
+      # should re-check if the manual smoke later reveals trouble.
+      - name: Test API Health (staging)
+        run: |
+          curl -f ${{ steps.frontend.outputs.api_url }}/v1/auth || echo "API not yet ready, may need warm-up"
+
       - name: Staging Deployment Summary
         run: |
           echo "========================================="
@@ -661,6 +699,15 @@ jobs:
     name: Deploy Backend to Production
     runs-on: ubuntu-latest
     needs: test
+    # Boy-scout (#253 expansion round): same `timeout-minutes` discipline as
+    # deploy-dev/deploy-staging above, with extra headroom. Prod runs the
+    # additional `CDK Diff (show changes for audit trail)` step (~30s) on top
+    # of the full deploy + frontend rebuild + CloudFront wait, and a prod
+    # CFN update can take longer than staging if there are more resources to
+    # update (Cognito user-pool changes, in particular, can take 5-10 min on
+    # their own). 45 min gives ~3x headroom over a typical full deploy while
+    # still tripping before the GitHub Actions 6-hour cap.
+    timeout-minutes: 45
     # Issue #160 (branch-guard asymmetry): mirrors the fix applied to
     # deploy-staging above. See that job's comment block for rationale.
     if: github.ref == 'refs/heads/main' && github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'prod'
@@ -750,6 +797,18 @@ jobs:
           stack-name: LfmtPocProd
           aws-region: ${{ env.AWS_REGION }}
           node-version: ${{ env.NODE_VERSION }}
+
+      # Boy-scout (#253 expansion round): mirror deploy-dev's `Test API Health`
+      # step (line 312) and the analogous staging step above. Same rationale:
+      # cheap audit signal for the CFN-succeeds-but-API-unreachable failure
+      # mode. Non-blocking by design (`|| echo …`) — a transient 5xx during
+      # custom-domain propagation should not fail the prod deploy run, but
+      # the line in the run log is the audit signal that lets the operator
+      # know to re-check via manual smoke if the production-frontend traffic
+      # later shows trouble.
+      - name: Test API Health (prod)
+        run: |
+          curl -f ${{ steps.frontend.outputs.api_url }}/v1/auth || echo "API not yet ready, may need warm-up"
 
       - name: Production Deployment Summary
         run: |
--- a/.github/workflows/deploy-backend.yml
+++ b/.github/workflows/deploy-backend.yml
@@ -579,6 +579,61 @@ jobs:
         working-directory: backend/infrastructure
         run: npm ci
 
+      # Issue #253 (Gap 1) — staging/prod deploy-pipeline parity:
+      # mirror deploy-dev's idempotent CDKToolkit guard. Without this, the
+      # first-ever staging deploy against a fresh account/region fails with
+      # "This stack uses assets, so the toolkit stack must be deployed to
+      # the environment (Run `cdk bootstrap aws://...`)" — a footgun that
+      # only fires once but blocks the very first stand-up of a new staging
+      # account. The block is textually identical to the one in deploy-dev
+      # and deploy-prod (grep for `CDK Bootstrap (if needed)`); do not
+      # refactor into a composite action without keeping all three call-
+      # sites parallel — operators rely on the inline form for run-log
+      # readability when triaging a fresh-account bootstrap.
+      - name: CDK Bootstrap (if needed)
+        working-directory: backend/infrastructure
+        run: |
+          if ! aws cloudformation describe-stacks --stack-name CDKToolkit --region ${{ env.AWS_REGION }} 2>/dev/null; then
+            echo "CDKToolkit not found, bootstrapping..."
+            ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+            npx cdk bootstrap aws://${ACCOUNT_ID}/${{ env.AWS_REGION }}
+          else
+            echo "CDKToolkit already exists, skipping bootstrap"
+          fi
+
+      # Issue #253 (Gap 2) — intentional asymmetry with deploy-dev:
+      #
+      # deploy-dev runs a `cdk diff` short-circuit (Issue #163, the
+      # `Check for CDK infrastructure changes` step + the `if:
+      # steps.cdk-diff.outputs.has_changes == 'true'` gate above its
+      # `CDK Deploy to Dev` step) that skips `cdk deploy` when no
+      # infrastructure changes are detected. deploy-staging and
+      # deploy-prod deliberately DO NOT replicate that optimization.
+      # Rationale:
+      #
+      #   1. Trigger semantics differ. deploy-dev fires on every push to
+      #      main AND on workflow_dispatch (default env=dev). The push-
+      #      triggered path runs on shared-types-only or backend-only
+      #      commits that may have no infra impact — that's the noop case
+      #      #163 optimizes away. deploy-staging/prod are workflow_dispatch
+      #      ONLY (see their job-level `if:` guards), so every invocation
+      #      carries explicit operator intent to deploy.
+      #
+      #   2. UX. A manual dispatch that exits in ~30s with "no changes" is
+      #      ambiguous: did the deploy run? Was the frontend rebuild step
+      #      gated out? Did the smoke check actually validate the new code?
+      #      Operators expect a manual staging/prod deploy to be a full
+      #      deploy, not a conditional fast-path.
+      #
+      #   3. Audit-trail-first design. deploy-prod already has a `CDK Diff
+      #      (show changes for audit trail)` step that records the proposed
+      #      change-set in the workflow log BEFORE the deploy runs. Adding
+      #      a has_changes gate would either duplicate that diff (wasted
+      #      ~30s) or remove the audit step (lost paper trail). Both are
+      #      worse than the current always-deploy behaviour.
+      #
+      # Do NOT replicate the deploy-dev short-circuit here without first
+      # re-evaluating points 1-3 against the then-current trigger semantics.
       - name: CDK Deploy to Staging
         working-directory: backend/infrastructure
         run: npx cdk deploy --context environment=staging --require-approval never
@@ -674,6 +729,14 @@ jobs:
         working-directory: backend/infrastructure
         run: npx cdk diff --context environment=prod || true
 
+      # Issue #253 (Gap 2) — intentional asymmetry with deploy-dev:
+      # see the matching block in deploy-staging (above the `CDK Deploy to
+      # Staging` step) for the full rationale. Short version: deploy-prod
+      # is workflow_dispatch-only, so every invocation is explicit operator
+      # intent — a has_changes gate would surprise operators and would
+      # either duplicate or supplant the audit-trail `cdk diff` step above.
+      # Keep this `cdk deploy` call unconditional. Do NOT add a has_changes
+      # gate without re-reading the rationale on deploy-staging.
       - name: CDK Deploy Main Infrastructure
         working-directory: backend/infrastructure
         run: npx cdk deploy LfmtPocProd --context environment=prod --require-approval never


leixiaoyu:
## OMC R1 Self-Review — Wave 2 Track F (deploy-pipeline staging/prod parity)

**Reviewed commit**: `1a84bc4` on `chore/issue-253-staging-prod-parity` (only commit; branch-cut from `939a5ca` on `main`).
**Diff base**: `939a5ca` (current `main` HEAD).
**Issues covered**: #253 — both gaps (Gap 1 code fix, Gap 2 design-decision + comment-block documentation).
**New tests vs base**: 0 (workflow-only change; existing `tests/workflows/test_deploy_path_filters.py` covers the structural contract and still passes).
**Reviewer note**: this PR modifies a deploy pipeline, not application code. Review is calibrated for operational-correctness, fresh-account bootstrap, and operator UX on a manual dispatch — not the usual code-quality lens.
**Date**: 2026-05-13

---

### Category pass/fail matrix

| Category | Result | Notes |
|---|---|---|
| Architecture / SOLID | **PASS** | The bootstrap-guard block is copied verbatim across dev/staging/prod (now all three identical). DRY temptation explicitly resisted — a composite-action refactor was considered and rejected because operators rely on the inline form when triaging fresh-account bootstrap failures in run logs (the alternative is a 2-click drill-down to the composite). Documented inline in the new staging block. |
| Operational correctness (fresh account) | **PASS** | Bootstrap guard's `if !` branch fires exactly when `describe-stacks` returns non-zero. `describe-stacks` on a missing stack returns exit 255 with a `ValidationError: Stack with id CDKToolkit does not exist` message — stderr suppression (`2>/dev/null`) keeps the workflow log clean. Mental-model trace below. |
| Operational correctness (existing account) | **PASS** | `describe-stacks` returns 0 on success → the `if !` branch is skipped → bootstrap is a no-op. Cost: one extra ~500ms CloudFormation API call per staging deploy (negligible). |
| Defense-in-depth | **PASS** | Three layers prevent a wrong-environment bootstrap: (a) job-level `if:` already gates staging on `github.ref == 'refs/heads/main' && environment == 'staging'`; (b) `--region ${{ env.AWS_REGION }}` is explicit on both `describe-stacks` and `cdk bootstrap`; (c) `aws://${ACCOUNT_ID}/${AWS_REGION}` is constructed from `aws sts get-caller-identity` — so bootstrap goes to whichever account OIDC actually landed in, not an off-by-one literal. |
| Design decision quality (Gap 2) | **PASS** | Three-point rationale documented inline (trigger semantics, manual-dispatch UX, audit-trail-first design). The decision is grep-able from the workflow itself, not buried in a PR description. |
| Test coverage | **PASS** | `test_deploy_path_filters.py` still passes — no path-filter regressions. No new test was warranted: the change is a guard + comment blocks, neither of which has a meaningful unit-test surface beyond what actionlint + the path-filter test already cover. |
| CI/CD parity | **PASS** | The new staging block now matches dev and prod textually. A diff between the three `CDK Bootstrap (if needed)` blocks would be 0 lines (modulo whitespace). |
| Documentation | **PASS** | Every defensive choice has an inline `# Issue #253` comment that records the WHY. Future contributors reading the file in isolation can see why the asymmetry is intentional. |
| PR alignment | **PASS** | Diff is exactly the two changes described in #253 — no scope creep, no boy-scout pickups (4 observed out-of-scope items listed in PR body for future tracking). |

**No Critical / No High findings.** 1 Medium observation (SC2086 actionlint warning, addressed via "verbatim-copy" instruction). 0 Low observations.

---

### Critical review by gap

#### Gap 1 — staging `CDK Bootstrap (if needed)` block

**Mental-model trace, fresh account (no CDKToolkit stack)**:
1. `aws cloudformation describe-stacks --stack-name CDKToolkit --region us-east-1 2>/dev/null` → exit 255, `ValidationError: Stack with id CDKToolkit does not exist` written to stderr → stderr discarded → bash sees exit 255 → `!` inverts → `if true; then` branch entered.
2. `ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)` → succeeds because OIDC role assumption already happened in the `Configure AWS credentials (OIDC)` step above.
3. `npx cdk bootstrap aws://${ACCOUNT_ID}/${AWS_REGION}` → succeeds (assuming the OIDC role has the needed IAM; same IAM that dev/prod use).
4. Subsequent `cdk deploy` step succeeds because CDKToolkit now exists.

**Mental-model trace, existing account (CDKToolkit present)**:
1. `aws cloudformation describe-stacks --stack-name CDKToolkit --region us-east-1 2>/dev/null` → exit 0, stack metadata printed to stdout (also fine to display in the run log).
2. `!` inverts → `if false; then` → `else` branch fires → `echo "CDKToolkit already exists, skipping bootstrap"` → done.
3. Next step proceeds. No bootstrap re-run, no side effect.

**Red-team — what if CDKToolkit exists but is mid-update (UPDATE_IN_PROGRESS)?**
`describe-stacks` still returns 0 for any non-terminal state including `UPDATE_IN_PROGRESS` and `ROLLBACK_IN_PROGRESS`. The guard skips bootstrap, and the subsequent `cdk deploy` step will fail when it tries to upload assets to a stack that's mid-update. **This is identical to dev/prod behaviour** — not a regression introduced by this PR, just an existing operational footgun that the same guard pattern carries. If a future contributor wants to harden this, the right fix is to filter on `StackStatus IN (CREATE_COMPLETE, UPDATE_COMPLETE, IMPORT_COMPLETE)` across all three guards simultaneously. Out of scope here.

**Red-team — what if the OIDC role lacks `cloudformation:DescribeStacks` permission?**
`describe-stacks` returns exit 254 with an `AccessDenied` error → stderr suppressed → bash sees non-zero → `!` inverts → bootstrap branch fires unnecessarily. **Bootstrap would then also fail** on the same IAM (bootstrap requires far more permissions than describe-stacks). The visible failure mode is `cdk bootstrap` exiting non-zero, which fails the step. Acceptable: a missing IAM is loudly diagnosed at the bootstrap call, not silently passed.

**Red-team — what if `${ env.AWS_REGION }}` resolves to an unexpected value?**
The `env:` block at line 43 hard-codes `AWS_REGION: us-east-1`. Any future change to make this configurable (e.g. matrix-per-region) would need to re-verify all three bootstrap blocks. The textually-identical pattern across dev/staging/prod makes this a single search-and-replace.

**Medium observation — SC2086 actionlint warning on the new block**:
`actionlint` reports `SC2086:info: Double quote to prevent globbing and word splitting` on the `aws://${ACCOUNT_ID}/${{ env.AWS_REGION }}` expansion (line 595 of the new staging block). This is identical to the pre-existing warning on the dev block (line 226 pre-change, 226 post-change) and prod block (line 664 pre-change, 719 post-change). The user's #253 brief explicitly instructed verbatim copy of the dev/prod block to staging; fixing the quoting here would diverge staging from dev/prod and violate that instruction. **Resolution**: copy fidelity wins. A separate boy-scout PR can add quotes to all three bootstrap blocks simultaneously if desired.

#### Gap 2 — design decision (keep staging/prod unconditional)

**Decision**: keep the `cdk deploy` call on `deploy-staging` and `deploy-prod` unconditional. Document the intentional asymmetry with deploy-dev via inline comment blocks on both jobs.

**Three-point rationale (matches PR body)**:
1. **Trigger semantics**: staging/prod are workflow_dispatch-only; every invocation carries explicit operator intent. The `#163` short-circuit is justified on the basis of push-triggered re-runs that aren't applicable here.
2. **UX**: a manual-dispatch exit in ~30s with "no changes" creates ambiguity ("did my deploy actually happen?") that the operator has to resolve via log-reading.
3. **Audit-trail-first**: deploy-prod already has a `CDK Diff (show changes for audit trail)` step that records the change-set BEFORE the deploy runs. Adding a has_changes gate would either duplicate this diff (wasted time) or replace it (lost audit log).

**Red-team — could the operator's ~30s noop-exit objection be solved differently?**
Yes — adding an `::notice::` annotation at the start of the deploy step would surface the deploy intent in the run summary regardless of fast-path. But this is *added* complexity to deliver an *uncertain* cost savings, against jobs that fire <10x/month. The simple "stay unconditional" answer is more robust.

**Red-team — could the audit-trail concern (point 3) be solved by adding `cdk diff` to staging?**
deploy-staging doesn't currently have a `cdk diff` step. Adding one is in scope for a future "audit-trail-on-staging" PR, but mixing it with the bootstrap fix here would blur this PR's scope. **Not introduced.**

**Red-team — does the asymmetry decision close out #253?**
Yes. The issue body lists "Decide: should staging/prod get the `#163` short-circuit? If no, add a comment in `deploy-staging` / `deploy-prod` documenting the intentional asymmetry." — both comment blocks are added with the full three-point rationale. `Closes #253` is the appropriate disposition.

---

### Items deliberately OUT of scope for this PR

| Item | Why deferred |
|---|---|
| Add `timeout-minutes` to `deploy-dev`/`deploy-staging`/`deploy-prod` | Real parity gap with `smoke-tests` and `integration-tests` (both got timeouts from PR #240). Out of scope for #253 — file as a separate issue for the next Wave 2 hygiene track. |
| Add `Verify AWS identity` step to `deploy-staging` | Present on dev + prod. Trivial fix but not in scope. |
| Add post-deploy API health-check / smoke-test to `deploy-staging` and `deploy-prod` | Larger architectural change: requires per-environment secret routing (e.g. `STAGING_USER_POOL_ID`). Not appropriate for a hygiene PR. |
| Hardening the SC2086 quoting across all three bootstrap blocks | Would diverge staging from dev/prod and violate the "verbatim copy" instruction in #253. Boy-scout PR if anyone wants to do all three at once. |
| Refactor the bootstrap block into a composite action | Considered and rejected — operators rely on the inline form for run-log readability when triaging fresh-account failures. Documented inline. |

All five items are documented in the PR body's "Out-of-scope parity gaps observed" section.

---

### Validation

| Check | Result |
|---|---|
| `python3 -c "import yaml; yaml.safe_load(...)"` | **PASS** |
| `python3 tests/workflows/test_deploy_path_filters.py` | **PASS** |
| `/tmp/actionlint .github/workflows/deploy-backend.yml` | 10 warnings total; 9 pre-existing on dev/prod (unchanged), 1 new on staging that matches the dev/prod pattern verbatim. No new warning *classes*. |
| Mental-model trace, fresh-account bootstrap | **PASS** (see Gap 1 critical review) |
| Mental-model trace, existing-account bootstrap | **PASS** (see Gap 1 critical review) |
| Comment blocks grep-able from the workflow | **PASS** (`grep -n "Issue #253" .github/workflows/deploy-backend.yml` returns 4 hits — Gap 1 block + Gap 2 staging block header + Gap 2 prod block header + cross-reference comment) |

---

### Diff summary

```
.github/workflows/deploy-backend.yml | 63 ++++++++++++++++++++++++++++++++++++
1 file changed, 63 insertions(+)
```

- +18 lines: staging `CDK Bootstrap (if needed)` block (10-line comment + 8-line step)
- +29 lines: staging Gap 2 comment block (above `CDK Deploy to Staging`)
- +8 lines: prod Gap 2 comment block (above `CDK Deploy Main Infrastructure`)
- +8 lines: blank-line / formatting consistency

Zero deletions. The change is purely additive; the existing dev/prod bootstrap blocks and the existing prod `cdk diff` audit step are unchanged.

---

**Ready for review.** No blockers. The Gap 2 design decision is the load-bearing review item — if the owner prefers Option B (extend the short-circuit), the change is mechanical (~15 lines, mirror the dev `cdk-diff` step + add a `if: steps.cdk-diff.outputs.has_changes == 'true'` on the deploy step) and could land in a follow-up PR.

leixiaoyu:
## OMC R2 Self-Review — Expansion Round (#253 fold-in pass)

**Reviewed commit**: `e284b69` on `chore/issue-253-staging-prod-parity`. Expansion-only review — covers changes ON TOP of `1a84bc4`. The OMC R1 review of `1a84bc4` remains authoritative for Gap 1 + Gap 2: https://github.com/leixiaoyu/lfmt-poc/pull/258#issuecomment-4446028578.
**Diff base**: `1a84bc4` (PR's first commit).
**Issues touched in this round**: #253 (expansion) + #260 (filed as deferred follow-up).
**New tests vs base**: 0 (still workflow-only; existing `tests/workflows/test_deploy_path_filters.py` still passes).
**Date**: 2026-05-13

---

### Category pass/fail matrix (expansion-only)

| Category | Result | Notes |
|---|---|---|
| Scope discipline | **PASS** | Owner asked to fold in only items that don't significantly grow scope; original item 4 (staging/prod smoke + integration) genuinely does and was deferred to #260 with explicit rationale. Items 1, 2, 3 are mechanical / additive / self-contained — exactly the fold-in criteria. |
| Architecture / SOLID | **PASS** | All five expansion-round additions are inline, grep-able by `Boy-scout (#253 expansion round)`, and cross-reference the dev-side step they mirror. No DRY refactor temptation: each block is a 1-step addition, not a sequence worth extracting. |
| Operational correctness — timeouts | **PASS** | Values picked from observed run times. deploy-dev noop run 25770555446 finished in 3m48s; 30 gives ~6x headroom. Staging full deploy estimated 10-13 min (CDK + frontend + CF wait); 30 gives ~2x. Prod gets 45 for Cognito user-pool changes which can run 5-10 min on their own. All three trip well before GitHub's 6-hour cap. Trade-off documented inline. |
| Operational correctness — staging identity verify | **PASS** | `aws sts get-caller-identity` is read-only, ~200ms, identical invocation to dev (line 209) and prod. No new permissions required — the OIDC role already has STS:GetCallerIdentity (it's an implicit permission of any assumed role). |
| Operational correctness — health-check curls | **PASS** | The curl uses `${{ steps.frontend.outputs.api_url }}` from the composite `rebuild-frontend` action which is already proven to populate that output on staging (via `LfmtPocStaging` stack lookup) and prod (`LfmtPocProd`). The `|| echo` non-blocking pattern is verbatim from dev. No new infrastructure dependencies. |
| Defense-in-depth | **PASS** | Timeouts add a NEW safety layer (job-level kill switch) on top of GitHub's 6-hour cap. Health check adds a NEW audit signal that triggers without failing the build (non-blocking by design). Identity verify adds a NEW pre-flight diagnostic. None of these add coupling between jobs. |
| Test coverage | **PASS** | actionlint: 40 warnings before commit e284b69, 40 after — verified via `git stash` round-trip. Path-filter test still passes. No new warning classes; the new curl steps consume `${{ steps.frontend.outputs.api_url }}` which is a GitHub Actions expression (resolved by the runner, not the shell), so no SC2086 is introduced. |
| Documentation | **PASS** | All five new blocks are tagged `Boy-scout (#253 expansion round)` so `grep -n "expansion round" .github/workflows/deploy-backend.yml` returns exactly the 5 additions. Each block explains WHY (the failure mode it catches) and WHY THIS VALUE (the wall-clock observation behind 30/45). |
| PR alignment | **PASS** | PR body rewritten to move items 1-3 of the original "Out of scope" list into resolved Gap 3/Gap 4/Gap 5 sections; item 4 moved to "Deliberately deferred" with rationale + #260 link. Body now accurately reflects the new scope. |
| Follow-up issue quality (#260) | **PASS** | #260 names the four sub-decisions a future implementer must make (secret routing, IAM trust, test-user isolation, failure-mode design) and gives an acceptance criteria checklist. Not just a TODO — a complete brief. |

**No Critical, No High findings. 0 Medium, 0 Low** observations on the expansion round. The expansion itself is genuinely lower-risk than the R1 work (Gap 2's design decision was the load-bearing item there; the expansion is mechanical hygiene).

---

### Critical review by item

**Item 1 (timeout-minutes on all three deploy jobs)**: Picked values based on observed run times rather than guessing. dev=30 is generous against the observed 4-min noop; staging=30 matches dev (no diff short-circuit means staging full-deploys are closer to a dev full-deploy in wall-clock); prod=45 accounts for the extra audit-diff step and Cognito user-pool update timings. **Red-team**: what if a real staging deploy regularly takes >25 min? The timeout would prematurely fail. **Mitigation**: the rationale comment block on each timeout records the expected wall-clock so a future operator who sees a timeout-induced failure has the calibration data in-file; bumping the value is a 1-line edit. The alternative (no timeout) is strictly worse — a hung CFN update would hold the deploy-${ref} concurrency lock for 6 hours.

**Item 2 (Verify AWS identity on staging)**: Trivial. `aws sts get-caller-identity` is the dev/prod step verbatim. **Red-team**: could the step itself fail and abort a staging deploy that would otherwise succeed? Only if OIDC credentials weren't successfully configured — in which case every subsequent AWS CLI call (the bootstrap check, `cdk deploy`, the CFN describe-stacks downstream) would also fail. Fast-failing here gives a clearer error than fast-failing 10 steps later.

**Item 3 (health-check curls on staging/prod)**: Mirrors dev's non-blocking pattern. **Red-team — could the curl ever silently mask a real deploy failure?** No: it's gated AFTER the rebuild-frontend composite, which itself fails the job on a non-zero exit. By the time the health check runs, CDK + S3 sync + CF invalidation have all already succeeded. The curl's purpose is purely to surface "yes the API is reachable from a clean network namespace" as a row in the run log — not to gate the build. The `|| echo` makes that purpose explicit. **Red-team — what if the API truly is down?** The next deploy will see the same curl line; the operator's manual smoke (PR #211 policy) will catch it. No regression vs the current behaviour (which has zero post-deploy verification on staging/prod beyond the rebuild-frontend exit status).

**Item 4 (deferred to #260)**: The decline rationale matches the owner's stated criteria — "decline if the change requires non-trivial new design, new secrets, new IAM, or per-environment branching that introduces meaningful risk." All four apply here. #260 captures the four sub-decisions a future implementer must make.

---

### Items still OUT of scope after this expansion

| Item | Why deferred |
|---|---|
| #260 — staging/prod smoke + integration | Tracked separately; requires per-env secret routing, IAM trust updates, Cognito test-user isolation policy. |
| Hardening SC2086 quoting across all three bootstrap blocks | Would diverge from dev/prod and violate the "verbatim copy" instruction from R1. Boy-scout PR if anyone wants to do all three. |
| Composite-action refactor of the bootstrap block | Considered and rejected in R1 (operators rely on inline form for run-log readability during fresh-account triage). |
| Adding `cdk diff` audit step to staging | Separate question — referenced in R1, not in scope here. |

---

### Validation (expansion round)

| Check | Result |
|---|---|
| `python3 -c "import yaml; yaml.safe_load(...)"` | **PASS** |
| `python3 tests/workflows/test_deploy_path_filters.py` | **PASS** |
| `/tmp/actionlint .github/workflows/deploy-backend.yml` | 40 warnings before commit e284b69, 40 after (verified via `git stash` round-trip). Zero new warning classes. |
| Grep-ability of expansion changes | **PASS** — `grep -c "expansion round" .github/workflows/deploy-backend.yml` returns 5 (matches the 5 additive blocks) |

---

### Diff summary (expansion only, e284b69)

```
.github/workflows/deploy-backend.yml | 59 ++++++++++++++++++++++++++++++++++++
1 file changed, 59 insertions(+)
```

- +10 lines: deploy-dev `timeout-minutes: 30` + rationale comment
- +7 lines: deploy-staging `timeout-minutes: 30` + rationale comment
- +9 lines: deploy-prod `timeout-minutes: 45` + rationale comment
- +8 lines: deploy-staging `Verify AWS identity` step + rationale
- +13 lines: deploy-staging `Test API Health (staging)` step + rationale
- +12 lines: deploy-prod `Test API Health (prod)` step + rationale

Zero deletions. Pure addition. Existing dev pipeline unchanged. The R1 Gap 1 + Gap 2 blocks from commit 1a84bc4 are unchanged.

---

**Ready for re-review.** No new load-bearing decisions in this round — the expansion is mechanical hygiene that the owner explicitly green-lit folding in. The R1 Gap 2 design decision (keep staging/prod unconditional) remains the only item that meaningfully needs owner sign-off; the expansion items are convergent best-practice additions.

xlei-raymond (approved):
As a Senior Reviewer, I have conducted an audit of PR #258. This PR is an excellent example of Operational Parity and Defensive Infrastructure. By standardizing our deployment pipelines across all environments, the team has eliminated several "first-run"
  footguns and established a consistent audit trail that is critical for production reliability.

  The execution here is highly professional, particularly in its documentation of non-obvious design decisions.

  ---

  1. Architectural Integrity: The Parity Principle
  The addition of the CDK Bootstrap (if needed) guard to the staging pipeline (Gap 1) is a critical "Day 0" fix.

  Educational Note:
  Infrastructure-as-Code (IaC) often suffers from the "Works on my Machine" (or "Works in Dev") syndrome. A common failure mode is a pipeline that assumes an environment is already prepared.
   * The Lesson: Design for Idempotency. By including the bootstrap guard in every environment's deploy job, we ensure that the pipeline is capable of standing up a fresh account or region without manual intervention. This is a foundational requirement for
     Disaster Recovery and multi-region expansion.

  2. Principled Design: Intentional Asymmetry
  I am particularly impressed by the decision not to extend the cdk diff short-circuit to staging and production (Gap 2), and the thoroughness of the documenting comments.

  Critique:
  A junior engineer might have optimized for "Speed" by copying the short-circuit logic to all jobs. However, the author correctly identified that Operator Intent and Audit Confidence outweigh the 5-minute saving on a manual dispatch.
   * The Lesson: Context matters. Automated push-triggered deploys (dev) should be optimized for speed (short-circuits). Manual operator-triggered deploys (staging/prod) should be optimized for Visibility and Verification. The UX of a manual "Deploy" button
     that finishes in 30s with "No Changes" is confusing; a full re-verification of the stack is the safer, more confidence-inspiring choice.

  3. Reliability: Concurrency and Budgets
  The addition of timeout-minutes to the deploy jobs is a foundational "Reliability" win.

  Mentorship Note:
  GitHub Actions' default 6-hour timeout is a "silent killer" in teams with shared concurrency groups. A single hung CloudFormation update could effectively lock the entire team's deployment pipeline for a whole afternoon.
   * The Lesson: Every job needs a Time Budget. By setting explicit 30/45-minute caps, the team ensures that a platform failure (AWS) cannot paralyze the development lifecycle. This is a "Boy Scout" win that pays dividends during high-pressure incidents.

  4. Audit Trail: Verification at the Edge
  The inclusion of Verify AWS identity and the non-blocking Test API Health curls in staging and prod significantly improves our observability.

  Observation:
  Surfacing the OIDC-assumed role in the logs provides a fast triage signal for "Wrong Account" errors. The API health check, while non-blocking, captures the "CFN Succeeded but Gateway Broken" failure mode—one of the hardest bugs to find without a live test
  signal.

  ---

  Final Verdict:
  APPROVE.

  The PR is technically sound, well-documented, and demonstrates a mature understanding of CI/CD operations. The use of "Boy-scout" annotations for the expansion-round changes makes the audit trail easy to follow.

  Team: Take note of the Rationale blocks in the YAML. This is the "Gold Standard" for documenting Why a pipeline is designed a certain way, saving the next person from having to "git blame" their way to an understanding.

  Merge when ready.

Issue #276, opened 2026-05-17
ux(auth): RegisterPage auto-login fallback shows misleading 'check your email' message for any failure cause


leixiaoyu:
## Summary

`RegisterPage.tsx` performs an auto-login after a successful registration (issue #222 pattern). When the auto-login fails, the `catch {}` block swallows the actual error and redirects to `/login` with a hardcoded message `"Account created! Please sign in — check your email if verification is required."` — regardless of the actual failure cause.

This is misleading when the auto-login fails for a reason that has nothing to do with email verification (rate limit, server error, network outage). The user is told to check their email; in reality they just need to retry.

**Severity: Medium** — narrow window (only fires when register succeeds but auto-login fails before tokens are stored), but the message is actively wrong and the user has no recovery signal.

## Symptom

Real-world scenarios where this misfires:

1. **TooManyRequestsException on the login leg**: User registers, but Cognito's rate limiter (per IP, per minute) trips on the immediate follow-up `InitiateAuth` call. `login.ts:182-194` returns 429 `"Too many login attempts. Please try again later."`. The user lands on `/login` with `"Account created! Please sign in — check your email if verification is required."` — neither correct nor actionable.
2. **Transient 500 from `login.ts` catch-all (line 197-208)**: Same misleading message.
3. **Network error between register-success and login-attempt**: Same.

In all cases the user's account exists (registration succeeded) but the on-screen message implies they need to check their email — which is irrelevant in dev (auto-confirm enabled) and only sometimes relevant in prod.

## Reproduction

1. `cd frontend && VITE_MOCK_API=true npm run dev`
2. Patch the MSW handler for `POST /auth/login` to return 429 once after the next `POST /auth/register`. (Or use the reserved-filename pattern from `frontend/LOCAL-TESTING.md` to inject the failure.)
3. Submit a valid registration form.
4. Observe: redirect to `/login` with the misleading message in the info `<Alert>`, not the actual `"Too many login attempts…"` from the login attempt.

## Root cause

`frontend/src/pages/RegisterPage.tsx:49-62`:

```ts
try {
  await login({ email: formData.email, password: formData.password });
  navigate(ROUTES.DASHBOARD);
} catch {
  // Auto-login failed (e.g., prod env where email verification is still
  // required, or a transient network error). Redirect to /login and pass
  // a friendly hint so the user knows their account was created.
  navigate(ROUTES.LOGIN, {
    state: {
      message:
[...]

Issue #274, opened 2026-05-17
ux(auth): LoginForm/RegisterForm/ForgotPasswordForm bypass getApiErrorMessage — leak raw 'Network Error' string


leixiaoyu:
## Summary

Auth forms (`LoginForm`, `RegisterForm`, `ForgotPasswordForm`) extract their submit-error text with a raw `err.message ?? "An error occurred..."` cast, bypassing the curated `getApiErrorMessage` extractor introduced in PR #268 (issue #266). This is the same antipattern that surfaced "An unexpected error occurred" on `TranslationDetail` — except on the auth pages it leaks the raw axios string (e.g. `"Network Error"`) instead of the curated `NETWORK_MESSAGE`, and silently swallows `STATUS_MESSAGES`/`COPY_BY_CODE` precedence.

**Severity: High** — common, user-visible flow (logging in / registering with a flaky network shows `"Network Error"` instead of "Connection lost — check your internet and try again.").

## Symptom

When the browser is offline (or the API endpoint is unreachable):

- `LoginForm` renders `<Alert>Network Error</Alert>` (raw axios `error.message` passed through `api.ts` response-error interceptor at line 401-407).
- `RegisterForm` renders the same.
- `ForgotPasswordForm` renders the same.

When the backend rate-limits a register call with HTTP 429 + `{ message: "Too many requests" }`, the user sees `"Too many requests"` (acceptable today) — but the same code path will leak any future generic backend `message` because no deny-list (`GENERIC_MESSAGES`) filter runs.

## Reproduction

1. `cd frontend && VITE_MOCK_API=true npm run dev`
2. Navigate to `/login`. Use the reserved filename or temporarily disable the dev server's `/auth/login` handler to force an axios `ERR_NETWORK`.
3. Submit the form. Expected: `"Connection lost — check your internet and try again."` (the `NETWORK_MESSAGE` constant from `translationErrorMessages.ts:79`). Actual: `"Network Error"`.

## Root cause

Three sibling sites in `frontend/src/components/Auth/`:

- `LoginForm.tsx:91-94`
  ```ts
  } catch (error) {
    const apiError = error as { message?: string };
    setSubmitError(apiError.message || 'An error occurred during login');
  }
  ```
- `RegisterForm.tsx:173-179` — same shape with `"An error occurred during registration"`.
- `ForgotPasswordForm.tsx:90-93` — same shape with `"An error occurred while requesting password reset"`.

[...]

PR #272, opened 2026-05-17
tech-debt(ui): sweep page error-render paths through getApiErrorMessage (#271)

leixiaoyu:
## Summary

Follow-up to #269 / PR #270. Four sibling sites in `frontend/src/pages/` surfaced raw `err.message` (lifted from `TranslationServiceError`) on error-render branches, bypassing the API-envelope precedence chain (`GENERIC_MESSAGES` filter → `response.data.message` → `COPY_BY_CODE` → `STATUS_MESSAGES` → fallback). This PR routes all four through `getApiErrorMessage`.

## PR #270 coordination

PR #270 merged at 2026-05-17T17:07:23Z, mid-flight on this branch. Scenario shifted (b) → (a): I committed my changes, then `git merge origin/main --no-edit` auto-merged PR #270's modifications to `TranslationDetail.tsx` (lines 302-322) and `TranslationDetail.test.tsx`. No conflicts; my in-page-alert change (line 138-143) lives at a distinct line range from PR #270's fatal-load early-return change. Both fixes coexist.

## Per-site change summary

| # | Site | Before | After |
| - | - | - | - |
| 1 | `TranslationDetail.tsx` `queryErrorMessage` derivation (in-page alert) | `queryError.message` | `getApiErrorMessage(queryError)` |
| 2a | `TranslationCompare.tsx` `fetchError` non-404/non-403 fallthrough (audit find) | `jobError.message` | `getApiErrorMessage(jobError)` — 404/403 hardcodes preserved |
| 2b | `TranslationCompare.tsx` blob-download catch | `err.message` | `getApiErrorMessage(err)` |
| 3 | `TranslationHistory.tsx` history-list load catch | `err.message` | `getApiErrorMessage(err)` |
| 4 | `TranslationHistory.tsx` history-download catch | `err.message` | `getApiErrorMessage(err)` (Alert at top of page IS user-visible) |

## Recursive audit findings

Beyond the 4 sites enumerated in #271, I found one trivial sibling in scope and folded it in:

- **`TranslationCompare.tsx:59`** — the `fetchError` derivation's non-404/non-403 fall-through returned `jobError.message` directly. Same antipattern, same fix, same file. Folded in (logged as 2a above).

Out-of-scope (filed for future):
- `FileUploadForm.tsx:185` — upload-flow error toast (different UX path).
- `TranslationProgress.tsx:67,84` — component-level (not a page).

[...]

PR #283, opened 2026-05-17
fix(api): preserve backend message on 403 (#275)

### LINKED ISSUE #275: ux(auth): api.ts 403 interceptor clobbers backend 'verify your email' prose for unconfirmed-user login
## Summary

The axios response-error interceptor in `frontend/src/utils/api.ts` rewrites EVERY HTTP 403 response's `message` to the generic `ERROR_MESSAGES.UNAUTHORIZED` constant (`"You do not have permission to perform this action."`). This silently destroys backend-emitted user-facing prose for the auth path — most importantly the email-verification guidance returned by `login.ts` for `UserNotConfirmedException`.

**Severity: High** — affects every user with an unconfirmed email address in production. They see "You do not have permission to perform this action." instead of "Please verify your email address before logging in. Check your inbox for the verification link." — the latter has the recovery action; the former does not.

## Symptom

Production user with an unverified email tries to log in. Backend `login.ts:167-179` returns:

```
HTTP 403
{
  "message": "Please verify your email address before logging in. Check your inbox for the verification link.",
  "requestId": "<uuid>"
}
```

`LoginForm` renders an `<Alert severity="error">` containing:

```
You do not have permission to perform this action.
```

The user has no idea their email needs verification.

## Reproduction

In a production-like environment (dev defaults to `autoConfirmUser = true` so this code path is unreachable):

1. Register a new user against a Cognito user pool with `autoVerify` disabled OR temporarily disable the PreSignUp Lambda's `autoConfirmUser` flag.
2. Without confirming the email, navigate to `/login` and submit valid credentials.
3. Observe `<Alert>You do not have permission to perform this action.</Alert>` — should be the prose from `login.ts:175`.

Alternative reproduction via the unit-test layer: any test that mocks `apiClient.post('/auth/login', ...)` to reject with `{ status: 403, response: { data: { message: "Please verify your email…" } } }` and expects the rendered alert to contain the verification prose will fail.

## Root cause

`frontend/src/utils/api.ts:389-398`:

```ts
// Handle 403 Forbidden
if (axiosError.response?.status === 403) {
  const apiError: ApiError = {
    message: ERROR_MESSAGES.UNAUTHORIZED,                       // <-- overrides backend prose
    status: 403,
    data: axiosError.response.data,
  };
  return Promise.reject(apiError);
}
```

The 403 branch unconditionally replaces the message. The 400/422 branch (line 410-419) and the generic branch (line 433-442) BOTH preserve `axiosError.response.data?.message` when present. Only 403 has this drift.

## Acceptance criteria

- [ ] `api.ts:389-398` is updated to preserve `axiosError.response.data?.message` when present, falling back to `ERROR_MESSAGES.UNAUTHORIZED` only when absent.
- [ ] The rewrite matches the 400/422 pattern (`backendMessage || ERROR_MESSAGES.UNAUTHORIZED`).
- [ ] Unit test in `api.test.ts`: 403 with body `{ message: "Please verify your email…" }` surfaces that prose; 403 with empty body surfaces `ERROR_MESSAGES.UNAUTHORIZED`.
- [ ] No regression on existing 403 paths that have no body (e.g. API Gateway authorizer 403s — these have no `data.message`).
- [ ] The forgot-password / register flows don't need any change; they don't currently emit 403.

## Why this is safe

Backend 403 emitters reviewed:
- `login.ts:167-179` — emits the email-verification prose. **Currently lost — this issue restores it.**
- API Gateway Cognito authorizer 403 (on protected endpoints with missing/expired tokens) — emits an opaque body that does not contain `data.message`. The fallback (`|| ERROR_MESSAGES.UNAUTHORIZED`) covers this.
- `getCurrentUser.ts:48-55` — emits 401, not 403; unaffected.

There is no current path that emits 403 with sensitive backend prose worth suppressing. If a future endpoint adds one, the right fix is at that endpoint, not in the universal interceptor.

## References

- `frontend/src/utils/api.ts:389-398` (offending branch)
- `backend/functions/auth/login.ts:167-179` (UserNotConfirmedException → 403)
- PR #251 — established the precedence pattern that this issue extends to 403.

## Out of scope

- Adding more verification-flow UX (resend-link button, etc.) — separate issue once the message reaches the user.
- Auth form error extractor refactor (filed separately as the `getApiErrorMessage` sweep issue).

## Severity

**High** — direct production bug for any unconfirmed-email user (a guaranteed segment if email verification is ever enabled in prod). The current "You do not have permission to perform this action" is misleading and removes the only recovery path the backend provides.

leixiaoyu:
## Summary

The axios response-error interceptor's 403 branch in `frontend/src/utils/api.ts` unconditionally overwrote the backend's `message` with the generic `ERROR_MESSAGES.UNAUTHORIZED` constant ("You do not have permission to perform this action."). This silently destroyed user-facing prose emitted by `backend/functions/auth/login.ts:175` for `UserNotConfirmedException`, so unconfirmed-email users saw the generic copy instead of "Please verify your email address before logging in. Check your inbox for the verification link." — losing the only recovery action the backend provides.

This PR introduces an `extractBackendMessage` helper (in `frontend/src/utils/api.ts`) that mirrors the precedence rule from `getApiErrorMessage` in `frontend/src/utils/translationErrorMessages.ts`: surface the backend message verbatim when it is specific; fall back to the curated `ERROR_MESSAGES.UNAUTHORIZED` constant when the message is absent, empty/whitespace, on a non-object body, or in the `GENERIC_BACKEND_MESSAGES` deny-list (which guards against opaque strings like "Forbidden" or "Request failed" re-introducing the issue #266 bug class).

## Helper function

- **Name**: `extractBackendMessage(data: unknown): string | undefined`
- **Location**: `frontend/src/utils/api.ts` (module-private, alongside the `GENERIC_BACKEND_MESSAGES` deny-list it consults)

The helper is intentionally inlined rather than imported from `translationErrorMessages.ts`. Cross-file edits are minimized, the deny-list is self-contained in the file that uses it, and pulling api.ts into `translationErrorMessages.ts`'s dependency graph would risk a circular import (the latter already dispatches API errors via `getApiErrorMessage` which itself references the envelope shape).

## Composition with Stream P (#274/#277/#278/#279)

Stream P's PR will modify the auth forms + `AuthContext` to route caught errors through `getApiErrorMessage` instead of raw `err.message`. The two changes COMPOSE:

- **Pre-#275, pre-Stream P**: The interceptor clobbered the 403 message → the form rendered the generic UNAUTHORIZED constant.
- **Post-#275, pre-Stream P**: The interceptor surfaces the backend's verbatim prose → the form renders the correct verification message via raw `err.message`.
- **Post-#275, post-Stream P**: The interceptor surfaces backend prose → the form renders it via `getApiErrorMessage`, which has the same precedence rule the interceptor now mirrors. Identical user-visible outcome.

No file overlap with Stream P (`AuthContext.tsx`, `LoginForm.tsx`, `RegisterForm.tsx`, `ForgotPasswordForm.tsx`). No file overlap with Stream R (#276, `RegisterPage.tsx`).

## Test counts before/after

| Workspace | Before (file-count / passed) | After (file-count / passed) |
| --- | --- | --- |
| frontend | 38 / 795 | 38 / 800 |
| backend/functions | 30 / 626 | 30 / 626 |
| backend/infrastructure | 1 / 91 | 1 / 91 |
| shared-types | 2 / 24 | 2 / 24 |

Frontend coverage post-change: **97.96% statements / 93.47% branches** (api.ts: 99.1% / 93.61%).

Net delta: **+5 frontend unit tests** (one pre-existing 403 test rewritten as the deny-list case; five new cases for the specific-message, empty-body, whitespace-only-message, non-object-body, and case-insensitive-deny-list scenarios).

## Backend 403 prose audit (security check)

I walked every backend handler that returns a 403 to verify the prose is safe to surface verbatim:

| Handler | 403 prose | Safe to surface? |
| --- | --- | --- |
| `backend/functions/auth/login.ts:175` | "Please verify your email address before logging in. Check your inbox for the verification link." | **Yes** — generic guidance, no PII or resource-existence leak. This is the load-bearing case for #275. |
| `backend/functions/jobs/startTranslation.ts:130` | "You do not have permission to translate this job" + `errorCode: 'FORBIDDEN'` | **Pre-existing concern, NOT introduced by this PR.** The handler returns 404 first when the job doesn't exist (line 117-125), so a 403 confirms the job exists but isn't owned by the caller — a mild resource-existence leak. This is a **backend** issue (the prose itself isn't more revealing than the status code), not a frontend issue. Filing as out-of-scope follow-up — see below. |

No other backend handler in `backend/functions/**` returns a 403 (verified by grep of all `createErrorResponse(403` / `, 403,` patterns).

## Sibling-status-code overrides (out-of-scope finding)

While scoping this fix I noticed `api.ts:423-430` (the 500+ branch) ALSO replaces the backend message with the curated `ERROR_MESSAGES.SERVER_ERROR` constant. This is the same bug class as #275 but per the issue's "Out of scope" section the 403-only scope is preserved. Per the agent instructions:

> If you find a sibling 401/500 override that has the same bug class, file a follow-up issue rather than expanding scope.

A backend-side reaffirmation: the issue's "Why this is safe" section observes that the 500-class API contract dictates a human-readable phrase only. If the future backend signal warrants surfacing, a follow-up should:

1. Adopt the same `extractBackendMessage` helper (now exported-ready in api.ts; trivially promotable).
2. Add unit tests mirroring the 403 ones in this PR.

The 401 branch is intentionally different (the curated SESSION_EXPIRED copy must always win on terminal 401, regardless of backend prose, because the session is dead). No change recommended there.

I will file the 500+ follow-up as a separate issue, NOT in this PR.

## Out of scope

- Stream P: auth forms + AuthContext error extraction (#274/#277/#278/#279)
- Stream R: RegisterPage navigation (#276)
- `startTranslation.ts:130` existence-leak (separate backend concern)
- `api.ts:423-430` 500-class bug-class match (separate follow-up)

## Test plan

- [x] `npm ci && npx tsc --noEmit` (cold + warm) — frontend
- [x] `npm run lint` — frontend
- [x] `npx prettier --check .` — frontend
- [x] `npx vitest --run --coverage` — frontend (CI-parity command)
- [x] `npm test` — backend/functions
- [x] `npm test` — backend/infrastructure
- [x] `npm test` — shared-types

Closes #275

### DIFF
--- a/frontend/src/utils/__tests__/api.test.ts
+++ b/frontend/src/utils/__tests__/api.test.ts
@@ -583,11 +583,70 @@ describe('API Client - Response Error Interceptor', () => {
     }
   });
 
-  it('should handle 403 Forbidden', async () => {
+  // Issue #275: 403 backend-message precedence.
+  //
+  // Pre-#275 the 403 branch clobbered every backend message with the
+  // generic `ERROR_MESSAGES.UNAUTHORIZED` constant, silently destroying
+  // user-facing prose like the email-verification guidance emitted by
+  // `backend/functions/auth/login.ts:175` for `UserNotConfirmedException`.
+  // Post-#275 the precedence is: specific backend message wins; generic
+  // backend message OR absent backend message falls back to the
+  // `ERROR_MESSAGES.UNAUTHORIZED` constant (the deny-list guard prevents
+  // an opaque "Forbidden" / "Request failed" from re-introducing the
+  // issue #266 "An unexpected error occurred" class of bug).
+
+  it('403: surfaces specific backend message verbatim (issue #275 load-bearing case)', async () => {
     const { createApiClient } = await import('../api');
     const client = createApiClient();
 
-    // Mock adapter to simulate 403 error
+    // Mock adapter to simulate the 403 from login.ts:175 for an
+    // unconfirmed-user login. The pre-#275 implementation would
+    // clobber `error.message` with "You do not have permission to
+    // perform this action." — destroying the only recovery action
+    // the backend gave the user.
+    client.defaults.adapter = async () => {
+      const error: any = new Error('Request failed with status code 403');
+      error.isAxiosError = true;
+      error.response = {
+        status: 403,
+        data: {
+          message:
+            'Please verify your email address before logging in. Check your inbox for the verification link.',
+          requestId: 'req-abc',
+        },
+        statusText: 'Forbidden',
+        headers: {},
+        config: {},
+      };
+      throw error;
+    };
+
+    try {
+      await client.post('/auth/login', { email: 'x', password: 'y' });
+      expect.fail('Should have thrown an error');
+    } catch (error: any) {
+      expect(error.status).toBe(403);
+      expect(error.message).toBe(
+        'Please verify your email address before logging in. Check your inbox for the verification link.'
+      );
+      // The raw envelope is preserved for downstream consumers
+      // (e.g. `getApiErrorMessage` which also reads `data.errorCode`).
+      expect(error.data).toEqual({
+        message:
+          'Please verify your email address before logging in. Check your inbox for the verification link.',
+        requestId: 'req-abc',
+      });
+    }
+  });
+
+  it('403: falls back to UNAUTHORIZED constant when backend message is in the GENERIC deny-list', async () => {
+    const { createApiClient } = await import('../api');
+    const client = createApiClient();
+
+    // API Gateway authorizer 403 (no specific prose, just "Forbidden").
+    // Surfacing "Forbidden" verbatim would re-introduce the issue #266
+    // "An unexpected error occurred" class problem, so the deny-list
+    // routes us to the curated UNAUTHORIZED copy.
     client.defaults.adapter = async () => {
       const error: any = new Error('Request failed with status code 403');
       error.isAxiosError = true;
@@ -609,6 +668,135 @@ describe('API Client - Response Error Interceptor', () => {
       expect(error.message).toBe('You do not have permission to perform this action.');
     }
   });
+
+  it('403: falls back to UNAUTHORIZED constant when response body has no message field', async () => {
+    const { createApiClient } = await import('../api');
+    const client = createApiClient();
+
+    // Some 403 responses (e.g. opaque API Gateway 403s, raw blocked
+    // responses) carry no message field at all. Fall back to the
+    // curated copy.
+    client.defaults.adapter = async () => {
+      const error: any = new Error('Request failed with status code 403');
+      error.isAxiosError = true;
+      error.response = {
+        status: 403,
+        data: {},
+        statusText: 'Forbidden',
+        headers: {},
+        config: {},
+      };
+      throw error;
+    };
+
+    try {
+      await client.get('/admin');
+      expect.fail('Should have thrown an error');
+    } catch (error: any) {
+      expect(error.status).toBe(403);
+      expect(error.message).toBe('You do not have permission to perform this action.');
+    }
+  });
+
+  it('403: falls back to UNAUTHORIZED constant when backend message is whitespace only', async () => {
+    const { createApiClient } = await import('../api');
+    const client = createApiClient();
+
+    client.defaults.adapter = async () => {
+      const error: any = new Error('Request failed with status code 403');
+      error.isAxiosError = true;
+      error.response = {
+        status: 403,
+        data: { message: '   \t  \n' },
+        statusText: 'Forbidden',
+        headers: {},
+        config: {},
+      };
+      throw error;
+    };
+
+    try {
+      await client.get('/admin');
+      expect.fail('Should have thrown an error');
+    } catch (error: any) {
+      expect(error.status).toBe(403);
+      expect(error.message).toBe('You do not have permission to perform this action.');
+    }
+  });
+
+  it('403: falls back to UNAUTHORIZED constant when response body is not an object', async () => {
+    const { createApiClient } = await import('../api');
+    const client = createApiClient();
+
+    // Defensive: if some upstream layer returns a string body, the
+    // extractor must not treat that as a usable message.
+    client.defaults.adapter = async () => {
+      const error: any = new Error('Request failed with status code 403');
+      error.isAxiosError = true;
+      error.response = {
+        status: 403,
+        // axios may sometimes pass through plain-string error bodies
+        // when content-type isn't json. The extractor returns undefined
+        // for any non-object data.
+        data: 'Forbidden',
+        statusText: 'Forbidden',
+        headers: {},
+        config: {},
+      };
+      throw error;
+    };
+
+    try {
+      await client.get('/admin');
+      expect.fail('Should have thrown an error');
+    } catch (error: any) {
+      expect(error.status).toBe(403);
+      expect(error.message).toBe('You do not have permission to perform this action.');
+    }
+  });
+
+  it('403: deny-list match is case-insensitive (defense against backend casing drift)', async () => {
+    const { createApiClient } = await import('../api');
+    const client = createApiClient();
+
+    // Backend or API Gateway may send "FORBIDDEN" or "forbidden" depending
+    // on layer. The deny-list normalizes via .toLowerCase() so all variants
+    // fall back to the curated copy.
+    client.defaults.adapter = async () => {
+      const error: any = new Error('Request failed with status code 403');
+      error.isAxiosError = true;
+      error.response = {
+        status: 403,
+        data: { message: 'FORBIDDEN' },
+        statusText: 'Forbidden',
+        headers: {},
+        config: {},
+      };
+      throw error;
+    };
+
+    try {
+      await client.get('/admin');
+      expect.fail('Should have thrown an error');
+    } catch (error: any) {
+      expect(error.status).toBe(403);
+      expect(error.message).toBe('You do not have permission to perform this action.');
+    }
+  });
+
+  // 401-path regression guard:
+  //
+  // Issue #275's fix only changed the 403 branch — the 401 branch's
+  // behavior is intentionally DIFFERENT (when a refresh has already
+  // been attempted OR there is no refresh token, the curated
+  // SESSION_EXPIRED constant always wins, regardless of backend prose,
+  // because the session is dead). The real 401 path is exhaustively
+  // covered by `api.refresh.test.ts` (which uses axios-mock-adapter to
+  // properly attach axiosError.config so the refresh-flow guard fires).
+  // We deliberately do not duplicate that coverage here with a
+  // hand-constructed axios error — earlier attempts at this would
+  // either skip the 401 branch entirely (when axiosError.config is
+  // undefined) or accidentally pass for the wrong reason.
 
   it('should handle network errors (no response)', async () => {
     const { createApiClient } = await import('../api');
--- a/frontend/src/utils/api.ts
+++ b/frontend/src/utils/api.ts
@@ -25,6 +25,74 @@ export interface ApiError {
 }
 
 /**
+ * Generic message strings that are too vague to surface to the user.
+ *
+ * Mirror of the `GENERIC_MESSAGES` deny-list in
+ * `frontend/src/utils/translationErrorMessages.ts` (PR #251 / issue #266).
+ * The list is intentionally inlined rather than imported because:
+ *   - The translationErrorMessages module is auth-context-free and
+ *     pulling api.ts into its dependency graph would create a circular
+ *     dependency the wrapping `getApiErrorMessage` helper already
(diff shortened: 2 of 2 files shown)


leixiaoyu:
## OMC R1 self-review — PR #283 (fix(api): preserve backend message on 403)

### 1. Pass/fail matrix

| Dimension | Verdict | Evidence |
| --- | --- | --- |
| Architecture | PASS | Single-file fix + module-private helper. No new public surface; deny-list mirrors the precedence already established by `getApiErrorMessage` in `translationErrorMessages.ts`. Helper is a pure function (`unknown → string \| undefined`), no state, no side effects. |
| Type safety | PASS | `npx tsc --noEmit` clean cold + warm. No `any` introduced. Helper consumes `unknown` and narrows defensively (`typeof data === 'object'`, `typeof candidate === 'string'`). |
| Test coverage | PASS | 5 new test cases on the 403 branch + the pre-existing "Forbidden" case rewritten as the deny-list assertion. Frontend totals 38 files / 800 passed / 14 skipped (was 795 passed); statement coverage 97.96%, api.ts 99.1% / 93.61% branches. |
| Backward compat | PASS | The pre-existing 403 case with `data: { message: 'Forbidden' }` continues to surface `ERROR_MESSAGES.UNAUTHORIZED` thanks to the deny-list — no callers' assertions break. Logged-out / interceptor / 400 / 422 / 500 / 401-refresh tests all unchanged and passing. |
| Security | PASS (with one pre-existing backend concern surfaced) | See section 2. |
| Composition with Stream P | PASS | See section 3. |

### 2. Security red-team — does surfacing backend 403 prose leak resource-existence?

I walked every backend handler that returns a 403:

- **`auth/login.ts:175`**: Emits "Please verify your email address before logging in. Check your inbox for the verification link." Safe — generic guidance, no PII, no resource-existence implication. This is the load-bearing case that #275 exists to repair.
- **`jobs/startTranslation.ts:130`**: Emits "You do not have permission to translate this job" + `errorCode: 'FORBIDDEN'`. **Pre-existing concern, NOT introduced by this PR**: the handler returns 404 first when the job doesn't exist (line 117-125), so a 403 confirms the job exists but isn't owned by the caller — a mild existence leak. However, my PR doesn't change what the backend chooses to send; it just stops clobbering it. If the existence leak is a real concern, the fix is at the backend (collapse 404 + 403 into a uniform 404 for non-owners, per the same pattern `shared/jobRepository.ts:13` already recommends). Filed downstream awareness only; no scope change.
- **No other handler in `backend/functions/**` returns a 403** (verified by grep for `createErrorResponse(403` / `, 403,` / `status: 403`).

Verdict: the security posture is unchanged or improved. The deny-list specifically catches opaque API-Gateway-emitted "Forbidden" strings so they cannot leak as user-visible noise.

### 3. Composition with Stream P (#274/#277/#278/#279)

I traced the user-visible path for a 403 + unconfirmed-user case end-to-end:

| State | Interceptor produces | Form catch block path | User-visible message |
| --- | --- | --- | --- |
| Pre-#275, pre-Stream P | `{ status: 403, message: ERROR_MESSAGES.UNAUTHORIZED, data: { message: 'verify your email...' } }` | Raw `err.message` | "You do not have permission to perform this action." (BUG — the #275 symptom) |
| Post-#275, pre-Stream P | `{ status: 403, message: 'Please verify your email...', data: { message: 'Please verify your email...' } }` | Raw `err.message` | "Please verify your email address before logging in. Check your inbox for the verification link." (CORRECT — #275 fixed in isolation) |
| Post-#275, post-Stream P | Same as row above | `getApiErrorMessage(err)` | Same string — `getApiErrorMessage` reads `err.message` first (translationErrorMessages.ts:160-168), and that field now carries the correct prose. (CORRECT — both fixes compose cleanly) |
| Pre-#275, post-Stream P | `{ status: 403, message: UNAUTHORIZED, data: { message: 'verify your email...' } }` | `getApiErrorMessage(err)` | Stream P's helper reads `err.message` first → UNAUTHORIZED constant. The `data.message` envelope IS read at translationErrorMessages.ts:151-152 but only as a fallback when `err.message` is empty. So Stream P's fix in isolation would NOT have fixed #275. (CORRECT diagnosis — both fixes are required.) |

This is why the two issues were filed separately: each fix is independently necessary. Stream P's fix routes the form through the curated `getApiErrorMessage` rendering, but if the underlying `err.message` is clobbered, even `getApiErrorMessage` surfaces the wrong copy.

No file overlap with Stream P (`AuthContext.tsx`, `LoginForm.tsx`, `RegisterForm.tsx`, `ForgotPasswordForm.tsx`). No file overlap with Stream R (#276, `RegisterPage.tsx`). Safe to land in any order.

### 4. Items NOT in scope (explicit list to prevent scope creep)

- #274/#277/#278/#279 (Stream P — auth forms + AuthContext error extraction)
- #276 (Stream R — RegisterPage navigation)
- Backend 403 prose audit on `startTranslation.ts:130` (existence-leak) — pre-existing, separate concern if pursued
- `api.ts:423-430` 500+ branch — sibling bug class, filed as **#284** for follow-up
- 401 branch — intentionally different (terminal-401 must surface SESSION_EXPIRED regardless of backend prose; well-covered by `api.refresh.test.ts`)

### 5. CI verification evidence

\`\`\`
$ gh pr view 283 --json statusCheckRollup
\`\`\`

To be populated after CI completes. Will edit this comment with the final rollup once `Run Tests` and `Build Infrastructure` finish.

### 6. Lessons / notes for future sessions

- `axios.isAxiosError()` recognition: hand-constructed error objects with `isAxiosError = true` work for tests of branches that don't read `axiosError.config` (400/403/422/500/network) but DON'T work for the 401 branch (which needs `axiosError.config.url`). For 401 path testing, use `axios-mock-adapter` (see `api.refresh.test.ts`). I removed an over-engineered 401 regression-guard test that hit this; the real 401 path is exhaustively covered there.
- The `GENERIC_BACKEND_MESSAGES` deny-list in api.ts is a deliberate duplicate of the one in `translationErrorMessages.ts`. If a third caller appears, promote to a shared util in `frontend/src/utils/errorMessages.ts` or similar. Documented inline.

xlei-raymond (approved):
As a Senior Reviewer, I have audited PR #283. This PR is a critical fix for a "Blind Spot" in our API interceptor logic that was sabotaging our authentication recovery paths. By moving from an unconditional status-based override to an "Intelligent
  Extraction" pattern, the team has restored the backend's ability to guide users through confirmation-flow blockers.

  ---

  1. Architectural Integrity: Consistent Precedence
  The implementation of extractBackendMessage is an excellent architectural win.

  Educational Note:
  Interceptors often act as a "Hammer"—bluntly overwriting data based on status codes. While this provides a safety net, it can destroy Contextual Value.
   * The Lesson: Prefer Specificity. By mirroring the precedence rules established in PR #272 (getApiErrorMessage), the team has ensured that the interceptor and the page-level components now share a single, unified definition of what constitutes a
     "surfaceable message." This prevents the "Shadow Clobbering" that was hiding the email-verification guidance.

  2. Defensive Engineering: The Deny-List Shield
  I applaud the team for the intentional duplication of the GENERIC_BACKEND_MESSAGES deny-list.

  Mentorship Note:
  A junior approach might have been to import the list from translationErrorMessages.ts to be "DRY" (Don't Repeat Yourself).
   * The Lesson: Avoid Circular Coupling. The author correctly identified that importing from the high-level extractor into the low-level API client would create a circular dependency. The decision to duplicate the set with clear documentation is the correct
     senior-level trade-off for System Decoupling. Using a Set for O(1) lookups and case-insensitive matching further hardens the system against minor backend casing drift.

  3. Reliability: Regression Guarding the Boundary
  The addition of 5 new unit tests for the 403 branch is exemplary.

  Observation:
  Previous tests for 403 simply asserted that some unauthorized message was returned. The new tests verify the Negative Space—specifically proving that "Forbidden" (deny-listed) is not surfaced, but "Please verify your email" (specific) is. This level of rigor
  ensures that we don't re-introduce Issue #266 ("An unexpected error occurred") while fixing Issue #275.

  4. Technical Maturity: Pruning Over-Engineering
  I noticed the removal of the over-engineered 401 regression test in api.test.ts.

  Educational Note:
  Writing tests that require complex manual mock-object fabrication (mimicking Axios's internal config structures) leads to brittle, hard-to-maintain test suites.
   * The Lesson: Right Tool for the Job. Recognizing that axios-mock-adapter (already used in api.refresh.test.ts) is the superior way to test the 401 path—and deleting the manual attempt—shows a focus on Test Suite Quality over mere "test count."

  ---

  Final Verdict:
  APPROVE.

  The PR is technically sound, surgically targeted, and architecturally aligned with our high engineering standards. It successfully resolves a major UX blocker for unconfirmed users while maintaining our security posture.

  Team: Excellent work on the "Red-Team" audit of other 403 sites. This level of thoroughness is exactly what I expect in our hardening phase.

  Merge immediately.

